PR #208, opened 2014-07-30
Write modes with zero eigenvalues in the NMD file

toliwa:
In the method writeNMD in nmdfile.py, add the parameter zeros.
It controls writing zero eigenvale modes to the NMD file. It
can be very useful to visually inspect rigid-body motion vectors,
especially in the context of protein docking.

If zeros is **False** (by default), this function skips modes
with zero eigenvalues.

If zeros is **True**, modes with zero eigenvalues are
written out, their scaling factor being the square-root
of the inverse of the mode number times 0.0001. This provides
descending factors consistent with the NMA modes.

### DIFF
--- a/prody/dynamics/nmdfile.py
+++ b/prody/dynamics/nmdfile.py
@@ -337,13 +337,17 @@ def parseNMD(filename, type=None):
     return nma, ag
 
 
-def writeNMD(filename, modes, atoms):
+def writeNMD(filename, modes, atoms, zeros=False):
     """Return *filename* that contains *modes* and *atoms* data in NMD format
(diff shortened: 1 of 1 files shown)


jamesmkrieger:
We could use this if we use a keyword other than zeros, which already means something in python. We could just change this code to use Zeros with an upper case Z.

I'm not sure how meaningful/correct the rotation and translation modes that we get from e.g. ANM calculations are though.

SHZ66:
@jamesmkrieger are you talking about numpy.zeros? If so, it won't cause name conflicts because numpy.zeros is in another namespace. We didn't merge it because we weren't sure if scaling factor if handled appropriately.

PR #210, opened 2014-08-01
Pass zeros argument from calcANM to calcModes

toliwa:
Pass zeros argument from calcANM to calcModes in anm.py, the zeros argument of calcANM was previously ignored. This bugfix corrects it.


SHZ66:
Thank you so much. The issue has been fixed.

PR #218, opened 2014-11-24
Logger verbosity

sixpi:
It seems that the logger does not honor verbosity settings in the config. I think this change fixes that.


sixpi:
Would be great to get this merged.

Issue #250, opened 2015-10-03
Bug when parsing resnum with an insertion code to create an Atom Selection (PDB parser)


jmbromley:
The documentation says that you should be able to select residues with no insertion code by e.g. select('resnum 27_') and select a particular insertion code with e.g. select('resnum 27A').  However in both ProDy 1.5.1 and ProDy 1.6.1b0 the former selects seemingly random residues whilst the latter fails to select anything.  As a workaround the undocumented select('resnum 27 icode _') and select('resnum 27 icode A') can be used.  Example interactive session demonstrating these errors follows below.

    >>> import prody
    >>> print prody.__version__
    1.6.1
    >>> atoms = prody.parsePDB('4jo1')
    @> PDB file is found in working directory (4jo1.pdb.gz).
    @> 7252 atoms and 1 coordinate set(s) were parsed in 0.06s.
    >>> sel = atoms.select('chain L resnum 27 name CA')
    >>> sel.getResnums()
    array([27, 27, 27])
    >>> sel.getIcodes()
    array(['', 'A', 'B'], 
          dtype='|S1')
    >>> sel = atoms.select('chain L resnum 27_ name CA')
    >>> sel.getResnums()
    array([  2, 301, 302])
    >>> sel = atoms.select('chain L resnum 27A name CA')
    >>> sel.getResnums()
    Traceback (most recent call last):
      File "<stdin>", line 1, in <module>
    AttributeError: 'NoneType' object has no attribute 'getResnums'
    >>> sel = atoms.select('chain L resnum 27 icode _ name CA')
    >>> sel.getResnums()
    array([27])
    >>> sel.getIcodes()
    array([''], 
          dtype='|S1')
    >>> sel = atoms.select('chain L resnum 27 icode A name CA')
    >>> sel.getResnums()
    array([27])
    >>> sel.getIcodes()
    array(['A'], 
          dtype='|S1')

jmbromley:
Additionally it would be great if residue numbers with insertion codes were treated as first class citizens, so that e.g. select('resnum >= 27') would pull out 27, 27A, 27B, 28 ... while select('resnum >= 27A') would only pull out 27A, 27B, 28 ... and select('resnum >= 27B') would pull out only 27B, 28, ...

jmbromley:
Just had a quick look at the source code.  I'm pretty sure than line 2267 of select.py should read:

    value = token[:-1]

rather than (as it currently does):

    value = token[:1]

[...]

PR #251, opened 2015-10-03
Fix parsing resnum's with included insertion code for Selection Groups (issue #250)

jmbromley:
Bug fix for issue #250 

Hopefully fixes Selection Groups so that ‘resnum’ selection strings with included insertion codes work as advertised in the documentation.


SHZ66:
Thank you for the debugging. I think what you pointed out are correct. But again we cannot merge this request because of some error in during the check. I made a new request #291. Thank you so much.

Issue #252, opened 2015-10-05
Slice style ranges behave incorrectly when given null range


jmbromley:
Using the slice style range specification when creating atom Selection Groups gives wrong result if a 'null' range (i.e. start and stop the same) is specified.

Specifically:

    AtomGroup.select('resnum 5:5')

should return either None or an error, since in slice specification the range is only "up to but not including" the last value.  However it currently returns the same result as

    AtomGroup.select('resnum 5')

which is wrong.  Admittedly nobody would do this on purpose, but when programmatically creating selection strings it could happen and should be correctly dealt with.

PR #253, opened 2015-10-05
Fix issue #252

jmbromley:
Fix issue #252 Correctly throw SelectionError when a slice style range is specified in
a selection string with the same start and end value.


SHZ66:
It could be nothing but some check is failed so we cannot take the risk of merging this request. I made a similar fix to yours in #290. But still thank you so much for your input and fix!

PR #261, opened 2015-10-31
Some minor changes

SHZ66:
- fix the problem that the data type of gamma is changed after checkENMParameters()
- some formattings in bbenmtools.c
- add parenthesis to rtbtools.c to avoid warnings


cihankayacihan:
I fixed those issues already on my latest commit so those are unnecessary.

PR #263, opened 2015-11-10
Some bugs

wzmao:
* Fix `fetchPfamMSA` url problem.
* Fix kdtree warning information.
* Add 2 warning for DI and SCA calculation.
* Fix the `showDirectInfoMatrix` bug.


prody/proteins/pdbfile.py
```diff
@@ -162,6 +162,10 @@ def parsePDBStream(stream, **kwargs):
     elif model != 0:
         ag = AtomGroup(str(kwargs.get('title', 'Unknown')) + title_suffix)
         n_csets = 0
+    if "needfullname" in kwargs:
```
SHZ66: Hi Wenzhi,

Thanks for the bug fixes! Because of some check errors in the requested branch (though it could be nothing), we cannot take the risk of merging it. Therefore I am redoing all your fixes. I want to ask if you can give some explanation on what this argument is doing? A few sentences is okay. 

Thanks,
She

wzmao: This fix should not be added, it will cause some unexpected error while writing to a PDB file.

SHZ66: Okay. Thank you!

PR #271, opened 2016-02-14
dynamics/plotting.py: Fix color normalization exception in showOverlapTable()

gokceneraslan:
Matplotlib no longer has a `matplotlib.pyplot.normalize()` function, use `matplotlib.colors.Normalize` instead. Fixes `module object has no attribute normalize` error.


cihankayacihan:
Already updated with travis testa.

PR #292, opened 2016-05-11
Some fixes from Mao #263

SHZ66:
- Some fixes from Mao #263
- A slight change of datatype check in logger


SHZ66:
I closed this because the travis-ci check failed.

Issue #282, opened 2016-04-23
Backwards compatibility to Python 2.7.x


Pavellarcrc:
Hi

In the file ProDy/prody/database/pfam.py, line 74: 

request = urllib.request.Request('http://hmmer.janelia.org/search/hmmscan', enc_params)

the method has no attribute "request" in Python 2.11 (https://docs.python.org/2/library/urllib.html)

So its not compatible to the older version.

Regards,

Pedro

SHZ66:
Hi, the compatibility problem should be fixed. Please see the pull request #298.

PR #298, opened 2016-05-12
Fixes in pfam.py

SHZ66:
- Fix the compatibility of searchPfam with Python 2.7.11 (#282);
- Update the link of HMMER search service.


SHZ66:
Tested in Python 2.7 and 3.5.

Issue #268, opened 2015-12-16
Can't parse PSF


USER762341:
I'm having issues parsing this PSF file with parsePSF (psffile.py).

[psf.zip](https://github.com/prody/ProDy/files/64698/psf.zip)

Looks like a bad calculation in line 66: lines = psf.readlines(71 * (n_atoms + 5))

I'm not really sure what that expression is for, but it's returning a very low number of lines compared to how many lines are really in the file.

cihankayacihan:
can you share the error message, your python version and operating system version.

SHZ66:
Hi, the problem is that the provided psf file has more than 71 characters per line which usually is the case. I modified the code to fit such situations, so it should be fixed in #302.

PR #302, opened 2016-05-17
Fix parsePSF robustness (see #268)

SHZ66:


SHZ66:
In .psf file, the number of characters in one line might be bigger than 71. That will cause an error when using parsePSF().

PR #331, opened 2016-10-02
Integration with py3Dmol

dkoes:
If py3Dmol has been imported (e.g., within a Jupyter notebook, which is the only place it makes sense), then showProtein will render using py3Dmol instead of matplotlib.

For example, see:
http://mscbio2025.csb.pitt.edu/notes/prody1.slides.html#/43

### DIFF
--- a/prody/proteins/functions.py
+++ b/prody/proteins/functions.py
@@ -129,7 +129,9 @@ def showProtein(*atoms, **kwargs):
         import StringIO, py3Dmol
         from pdbfile import writePDBStream
         pdb = StringIO.StringIO()
-        writePDBStream(pdb, atoms)
+        
+        for mol in alist:
+            writePDBStream(pdb, mol)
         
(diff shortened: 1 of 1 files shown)


dkoes:
An example notebook can be found here:
http://nbviewer.jupyter.org/github/3dmol/3Dmol.js/blob/9c8915f68d38db62bd9600854716f42eda4aeb1f/py3Dmol/prody.ipynb

cihankayacihan:
I am not sure why but there is an issue related to Java in example notebook. I can not visualize the structures.

dkoes:
Since there is no Java, I do not think you are having issues with it.  What browser/OS are you using?

cihankayacihan:
I am sorry. I mean Javascript. I tried in Chrome Version 53.0.2785.116 (64-bit), Version 9.1.2 (11601.7.7).

Issue #329, opened 2016-09-24
.getMasses() function returns None


jamieAmacpherson:
I need to calculate the mass matrix of a pdb structure:
###### 

> > > from prody import *
> > > topology = "free.pdb"
> > > struct = parsePDB(topology)
> > > mass = structure.getMasses()
> > > print mass
> > > None
> > > ###### 
> > > 
> > > I can't find the source code for this function making it difficult to diagnose the problem.

SHZ66:
Hi, the mass is not recorded in PDB files, therefore ProDy calculates it based on the atom type. It wasn't stored when parsing the PDB thus caused this issue. It has been fixed in #337. Thanks for your input.

PR #337, opened 2016-11-14
Fix getMasses()

SHZ66:
#329


SHZ66:
Test failed at python 3, I will fix it.

SHZ66:
It passes the test under Python 2.7, 3.4, and 3.5, but somehow stuck at 2.6. I checked the log, it seems that the test cannot download a package and thus timeout. I think this shouldn't be a problem on my side.

PR #395, opened 2017-05-17
Some minor changes before the release

jamesmkrieger:
This a comment just so it doesn't say no comment. These changes don't really do much and are described in the respective commits anyway.

### DIFF
--- a/prody/dynamics/plotting.py
+++ b/prody/dynamics/plotting.py
@@ -211,7 +211,7 @@ def showProjection(ensemble, modes, *args, **kwargs):
       * 3 modes: :meth:`~mpl_toolkits.mplot3d.Axes3D.plot`"""
 
     import matplotlib.pyplot as plt
-
+    plt.figure()
     projection = calcProjection(ensemble, modes, kwargs.pop('rmsd', True))
 
     if projection.ndim == 1 or projection.shape[1] == 1:
--- a/prody/dynamics/analysis.py
+++ b/prody/dynamics/analysis.py
@@ -661,7 +661,6 @@ def calcPerturbResponse(model, atoms=None, repeats=100, **kwargs):
     returnFormat = returnFormat.lower()
 
     if len(matrix_array) == 1:
-        LOGGER.info('Output has been returned as a single matrix (an array).')
         return matrix_array.reshape(n_atoms,n_atoms)
     
     if returnFormat is 'both':


prody/dynamics/plotting.py
```diff
@@ -211,7 +211,7 @@ def showProjection(ensemble, modes, *args, **kwargs):
       * 3 modes: :meth:`~mpl_toolkits.mplot3d.Axes3D.plot`"""
 
     import matplotlib.pyplot as plt
-
+    plt.figure()
```
SHZ66: There could be some cases people want to plot the projection on top of some other plots. Always creating a new figure for showProjection disallows such manipulations. But I am aware that other plotting functions have this line, so my suggestion is that how about we make a new configuration called for example "new_figure", default to be `True`, and it controls if a new figure is created or not before plotting?

jamesmkrieger: Sounds good. Would you like me to do that?

SHZ66: Great. I'd love to do that.

jamesmkrieger: I think it depends also which plotting tools we use as to whether there should be an option on this. For example, pcolor makes a mess rather than nicely plotting on top. This is particularly horrible when you include color bars as these then get added next to each other.

SHZ66: Oh sorry I misread. Yeah go ahead. I can also do that if you don't know where to look at.

SHZ66: Yes, that's true. We can add this config only to those vertices plotting functions, e.g. plot().

SHZ66: I will merge this request so you can change it on the most updated version.

jamesmkrieger: I can search through and work out where to put the option. If you know already then maybe it's better for you to do it.

SHZ66: I will do it then. In case you need to find it later, you can start from a function called confProDy

jamesmkrieger: I was thinking of having it as a kwarg for each function. People can then decide each time they want to plot something whether they want a new plot or not. It shouldn't be an overall ProDy configuration thing.

SHZ66: Okay. I will then add a "new_fig" to each plotting function, default is True.

jamesmkrieger: Yes, all the vertices plotting functions.

PR #451, opened 2017-10-26
Added STAR file parsing and writing capabilities

jamesmkrieger:
STAR files are parsed into nested dictionaries and nested dictionaries can be written back out as STAR files.


jamesmkrieger:
The test file that revealed those bugs can now be parsed successfully so the bugs are fixed. Apologies for the failed attempts.

PR #503, opened 2018-02-15
Added daliSearch and calcAverageCrossCorr

fongchun:


fongchun:
There is still a dependency error. I am debugging on this.

PR #512, opened 2018-02-17
updated showSequenceTree to return tree and msa

jamesmkrieger:
Previously it just plotted the tree.


prody/proteins/blastpdb.py
```diff
@@ -1,50 +1,65 @@
...
+    results.  Sleep time is multiplied by 1.5 when results are not ready.  
+    *timeout* (default is 120 s) determines when to give up waiting for the results.
     """
 
     if sequence == 'runexample':
         sequence = ('ASFPVEILPFLYLGCAKDSTNLDVLEEFGIKYILNVTPNLPNLFENAGEFKYKQIPI'
                     'SDHWSQNLSQFFPEAISFIDEARGKNCGVLVHSLAGISRSVTVTVAYLMQKLNLSMN'
                     'DAYDIVKMKKSNISPNFNFMGQLLDFERTL')
+    elif isinstance(sequence, Atomic):
+        sequence = sequence.select('calpha and chain %s' % chain).getSequence()
```
SHZ66: `chain` is not defined but I guess you want it to be an argument. May I suggest not to preprocess it (`calpha` is okay). If one wants to blast the sequence of a particular chain, he can (and most likely will) select that chain. ProDy's `Selection` is straightforward and one fewer argument is better. Personally, I feel it is actually more elegant not to have `chain` as an argument because it is already taken care by passing a chain to `sequence`. Otherwise, I think this is a nice extension of the `sequence` argument to accept `Atomic` input.

prody/proteins/blastpdb.py
```diff
@@ -1,50 +1,65 @@
...
+    elif isinstance(sequence, Sequence):
+        sequence = str(sequence)
+    elif isinstance(sequence, str):
+        if len(sequence) == 4 or len(sequence) == 5:
+            ag = parsePDB(sequence)
+            sequence = ag.select('calpha and chain %s' % chain).getSequence()
     else:
-        try:
-            sequence = ''.join(sequence.split())
-            _ = sequence.isalpha()
```
SHZ66: I think this interesting piece of code is to remove spaces, tabs, line breakers from the sequence. This could be useful when the sequence is long because those characters are invisible and they can be easily left in the string when copying & pasting. Please consider adding this preprocessing.

prody/proteins/blastpdb.py
```diff
@@ -1,50 +1,65 @@
...
-        except AttributeError:
-            raise TypeError('sequence must be a string')
-        else:
-            if not _:
-                raise ValueError('not a valid protein sequence')
-    headers = {'User-agent': 'ProDy'}
+        raise TypeError('seq must be an atomic class, sequence class, or str not {0}'
```
SHZ66: The name of the argument is "sequence" instead of "seq".

prody/proteins/blastpdb.py
```diff
@@ -282,10 +297,12 @@ def showSequenceTree(hits):
             inp.write("\n")
     cmd = clustalw + " hits.fasta"
     os.system(cmd)
+    msa = parseMSA('hits.aln')
```
SHZ66: I am not sure but please make sure that clustalw always creates 'hit.aln' as one of its output files.

prody/proteins/blastpdb.py
```diff
@@ -269,8 +283,9 @@ def showSequenceTree(hits):
     """
     clustalw = which('clustalw')
     if clustalw is None:
-        print("The executable for clustalw does not exists, install or add clustalw to path.")
-        return
+        try: clustalw = which('clustalw2')
+        except: raise ValueError("The executable for clustalw does not exist, \
```
SHZ66: I don't know who wrote the original code but this is surely not ProDy's style. Nice catch! 
One suggestion is that `ValueError` might not be appropriate here. I would suggest either `EnvironmentError` or `IOerror`. In any case, `RuntimeError` is always an option. Please refer to the [documentation](https://docs.python.org/2/library/exceptions.html).

Additionally, I don't understand why "hits.fasta" can be read before calling the clustalw. Do you know?

prody/proteins/header.py
```diff
@@ -915,7 +915,7 @@ def isSheet(secstrs):
     torf = secstrs == 'E'
     return torf
 
-def assignSecstr(header, atoms, coil=False):
+def assignSecstr(header, atoms, coil=False, report=True):
```
SHZ66: If a user or a function wants to turn off the outputs, he/she/it can use `confProDy(debug='warning')` or `confProDy(debug='none')` to turn off the message and turn it back on after this code is run. Let's keep and stick to this ProDy style, otherwise, we should provide this argument for every function that prints out messages! (which is heavy work and unnecessary =) )

prody/proteins/pdbfile.py
```diff
@@ -174,7 +174,7 @@ def parsePDBStream(stream, **kwargs):
 
     :arg stream: Anything that implements the method ``readlines``
         (e.g. :class:`file`, buffer, stdin)"""
-
+    report = kwargs.get('report', True)
```
SHZ66: The same as above.

prody/proteins/blastpdb.py
```diff
@@ -235,15 +240,15 @@ def getParameters(self):
 
         return dict(self._param)
 
-    def getHits(self, percent_identity=90., percent_overlap=70., chain=False):
+    def getHits(self, percent_identity=0., percent_overlap=0., chain=False):
```
SHZ66: It is better not to change these default values for compatible reasons. Changing default values of a long-standing function could be a bad practice as it may change the results of the same code and it may frustrate the users. A most conserved point of view is that for the sake of compatibility, unless necessary, low-level and old functions should not be modified if the output may change. Let me know if the original values are not good.

prody/proteins/psiblast.py
```diff
@@ -29,19 +29,29 @@ def psiBlastRun(sequence='runexample', cycles=2, filename=None, **kwargs):
         default is 2
     :type cycles: int
     """
-    jobId = kwargs.get('jobId',None)
+    psithr = kwargs.get('psithr', 1.0e-3)
+    jobId = kwargs.get('previousjobid','')
+    selectedHits = kwargs.get('selectedHits','selected_hits.txt')
```
SHZ66: Try to keep the variable/argument names consistent. While the preferences of different languages are very different, [Python prefers to separate the words by underscores for variables and functions (but not for classes).](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names)

SHZ66: PEP8 is good Python style guide but surely it is not strict rules.

prody/proteins/blastpdb.py
jamesmkrieger: It does. We should probably also have a general clustalw function and then we'd have a prefix argument instead of 'hits'

prody/proteins/blastpdb.py
jamesmkrieger: It writes the file "hits.fasta" from the hits so that clustalw can read it in. I changed inp to f_out to make that clearer.

prody/proteins/blastpdb.py
jamesmkrieger: It's not the expected behaviour for new users. I was expecting that getHits would get all the hits unless I specified filters and I would think other people would too.

I've changed it back anyway.

prody/proteins/psiblast.py
jamesmkrieger: In this case, I am mostly using what EBI PSI-BLAST has. Generally ok.

prody/proteins/blastpdb.py
jamesmkrieger: I restored the middle line. The rest of checking and I assumed this was too.

prody/proteins/blastpdb.py
jamesmkrieger: fixed

prody/proteins/header.py
jamesmkrieger: I guess. I was thinking more in terms of what happens when we use higher level functions. Then we want the reporting of the lower level function to be turned off because we have other summary-level reporting instead. Generally reporting would be true to the extent set by confProDy when using the function by itself.

We can talk about this more but I feel we just need to add it to a function when we make another function that runs it multiple times.

prody/proteins/pdbfile.py
jamesmkrieger: same as above

prody/proteins/blastpdb.py
```diff
@@ -45,12 +45,19 @@ def blastPDB(sequence='runexample', filename=None, **kwargs):
...
         sequence = sequence.select('calpha and chain %s' % chain).getSequence()
     elif isinstance(sequence, Sequence):
         sequence = str(sequence)
     elif isinstance(sequence, str):
         if len(sequence) == 4 or len(sequence) == 5:
-            ag = parsePDB(sequence)
+            ag = parsePDB(sequence[:4])
+
+            if len(sequence) == 5:
+                chain = sequence[-1]
```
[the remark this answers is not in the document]
jamesmkrieger: fixed

prody/proteins/psiblast.py
jamesmkrieger: ProDy style is to use lowercase with uppercase for new words in functions. We can have underscores in variables though.

jamesmkrieger: I think blastpdb.py and psiblast.py now have this to the extent possible.

prody/proteins/blastpdb.py
```diff
@@ -1,50 +1,73 @@
...
+    import urllib2
 
+from prody.sequence import Sequence
+from prody.atomic import Atomic
+from prody.proteins.pdbfile import parsePDB
+
+__all__ = ['PDBBlastRecord', 'blastPDB', 'showSequenceTree']
 
-def blastPDB(sequence, filename=None, **kwargs):
+def blastPDB(sequence='runexample', filename=None, **kwargs):
```
SHZ66: Let's leave `sequence` as a required/positional argument. It can make it clearer that `sequence` is the only and crucial argument of this function without needing to read the source code or the document. It may also help the users understand the function of this function is to blast a sequence against the PDB. Leaving some positional arguments is generally (not always) a [good practice](https://stackoverflow.com/questions/7041752/any-reason-not-to-always-use-keyword-arguments).

PR #532, opened 2018-02-27
Fix to showAlignment so it doesn't print empty lines if indices is None

jamesmkrieger:


prody/proteins/compare.py
```diff
@@ -954,14 +954,15 @@ def compare(m1, m2):
         mappings.sort(compare, reverse=True)
     return mappings
 
-def mapChainByChain(atoms, chain, **kwargs):
+def mapChainByChain(atoms, ref, **kwargs):
```
SHZ66: `chain` is designed to be a single chain to be consistent with `mapOntoChain`, and usually when these functions are called, [there is an outer iteration that splits the reference into individual chains](http://prody.csb.pitt.edu/tutorials/ensemble_analysis/dimer.html?highlight=mapontochain#prepare-ensemble).

jamesmkrieger: It would be good to automate further and include that iteration in a function. This seemed like a good place to put it. If the user does provide an outer iteration so there are only chains then it still works as chain.getHierView().iterChains() returns the single chain.

SHZ66: Ok, I agree. We probably can change mapOntoChain to the same style later.

PR #534, opened 2018-02-27
buildEnsemble now takes alignment_list and alignments_list as kwargs

jamesmkrieger:
These are indexed for passing to the mapping function

### DIFF
--- a/prody/ensemble/functions.py
+++ b/prody/ensemble/functions.py
@@ -333,7 +333,8 @@ def alignPDBEnsemble(ensemble, suffix='_aligned', outdir='.', gzip=False):
     else:
         return output
 
-def buildPDBEnsemble(refpdb, PDBs, title='Unknown', labels=None, seqid=94, coverage=85, mapping_func=mapOntoChain, occupancy=None, unmapped=None):
+def buildPDBEnsemble(refpdb, PDBs, title='Unknown', labels=None, seqid=94, coverage=85, mapping_func=mapOntoChain, \
+                     occupancy=None, unmapped=None, **kwargs):
     """Builds a PDB ensemble from a given reference structure and a list of PDB structures. 
     Note that the reference structure should be included in the list as well.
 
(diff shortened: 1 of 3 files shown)


jamesmkrieger:
The conflict arises because I tried changing things and you'd already merged my previous commit, which wasn't right.

SHZ66:
Alright, take your time to solve the conflicts. I will merge it when you are done.

prody/ensemble/functions.py
```diff
@@ -401,7 +424,9 @@ def buildPDBEnsemble(refpdb, PDBs, title='Unknown', labels=None, seqid=94, cover
         for chain in refchains:
             mappings = mapping_func(pdb, chain,
                                     seqid=seqid,
-                                    coverage=coverage)
+                                    coverage=coverage,
+                                    alignment=alignment_list[i],
+                                    alignments=alignments_list[i])
```
SHZ66: I think hiding alignment and alignments in kwargs is a good idea because these arguments are not required for every mapping functions.

prody/ensemble/functions.py
```diff
@@ -362,18 +363,40 @@ def buildPDBEnsemble(refpdb, PDBs, title='Unknown', labels=None, seqid=94, cover
...
 
     if not isinstance(refpdb, (Chain, Segment, Selection, AtomGroup)):
         raise TypeError('Refpdb must be a Chain, Segment, Selection, or AtomGroup.')
     
     if labels is not None:
         if len(labels) != len(PDBs):
-            raise TypeError('Labels and PDBs must have the same lengths.')
+            raise TypeError('labels and PDBs must have the same lengths.')
+
+    alignment_list = kwargs.get('alignment_list', None)
```
SHZ66: This processing is not necessary if the user did not provide alignment_list. alignment_list can be wrapped into kwargs and given to mapping_func

SHZ66: mapping_func is designed to be generic to all mapping functions including user-defined ones, so it should be as simple as possible. Ideally, only the four arguments are necessary, reference, target, seqid, and coverage. All the extra arguments that are needed for variants of mapping functions should go to kwargs to provide the flexibility and keep the simplicity at the same time.

prody/ensemble/functions.py
```diff
@@ -362,18 +363,40 @@ def buildPDBEnsemble(refpdb, PDBs, title='Unknown', labels=None, seqid=94, cover
...
+    if alignments_list is None:
+        alignments_list = [None]* len(PDBs)
+    else:
+        if len(alignments_list) != len(PDBs):
+            raise TypeError('alignments_list and PDBs must have the same lengths.')
 
     # obtain the hierarhical view of the reference PDB
     refhv = refpdb.getHierView()
-    refchains = list(refhv)
+    refchains = refhv.iterChains()
```
SHZ66: this will raise an error in the next line:
`atoms = refchains[0]`

iterators cannot be indexed so I converted hierview to a chain list.

SHZ66:
I am closing this request for now. @jamesmkrieger please rebase your fork and see #534 for the changes that may fulfill the need of this request.

PR #543, opened 2018-03-03
 Fixed _getBiomoltrans to work with new PDBs too

jamesmkrieger:
For example, PDB 5CNK has multiple blocks with 'APPLY THE FOLLOWING TO CHAINS:' lines and then 'BIOMT' lines.


SHZ66:
Is the request ready to be merged?

jamesmkrieger:
It was and I forgot and added more things. It will be again when the checks are passed.

PR #568, opened 2018-03-15
 Associated PDBEnsemble with an MSA

SHZ66:
- Associated PDBEnsemble with an MSA and added `getSequence ` to `PDBConformation`.
- Added `__add__` operation to MSA class.
- Now MSA class accepts list or array of strings, as long as the sequences have equal length if `aligned=True`, or gaps will be added to the end if they don't and `aligned=False`.


SHZ66:
I will fix this.

SHZ66:
The Python 3 related issue was fixed. Probably needs a more robust way to determine the string type.

PR #570, opened 2018-03-15
Added tests for PDBEnsemble associated MSA

SHZ66:
- Improvement of `addCoordset`. If `degeneracy=True` and `coords` is `Atomic`, then `getCoords` instead of `getCoordsets` will be used to retrieve coordinates. **This will allow using `setACSIndex` to alter active coordinates when adding an atom with multiple coordinate sets.**
- `other._atoms` will be used as the new atoms when adding ensembles if `self._atoms` isn't available. The same goes with weights.
- Bug fixes to `showAlignment` and **now it assumes default labels if labels are not available**. 
- Changed the default title for `MSA.__add__` in order to be consistent with other classes, such as `Ensemble`.


SHZ66:
The error is related to Python3 unicode string. I will fix it.

SHZ66:
Problem solved.

PR #571, opened 2018-03-15
 fixes to plotting and subgrouping trees 

ajimoreno:


prody/dynamics/plotting.py
```diff
@@ -228,9 +228,11 @@ def showProjection(ensemble, modes, *args, **kwargs):
 
     import matplotlib.pyplot as plt
 
+    returnAx = kwargs.pop('returnAx',False)
+
     cmap = kwargs.pop('cmap', None)
 
-    if SETTINGS['auto_show']:
+    if kwargs.pop('new_fig',True):
```
SHZ66: We are now abandoning `new_fig` and control the behavior by `auto_show`. If `confProDy(auto_show=True)` then a new figure will also be created.

prody/dynamics/plotting.py
```diff
@@ -336,6 +338,10 @@ def showProjection(ensemble, modes, *args, **kwargs):
 
     if SETTINGS['auto_show']:
         showFigure()
+
+    if returnAx:
```
SHZ66: The axes is not returned because it can be simply retrieved via `matplotlib.pyplot.gca`. `show` is returned so that it is consistent with matplotlib.

prody/dynamics/plotting.py
jamesmkrieger: No, it doesn't work like that. I've tested it lots of times. If auto_show is False, you don't get anything until you put plt.show() and when you do you still get multiple figures all the time.

prody/dynamics/plotting.py
jamesmkrieger: ok fine. I'll fix that then.

jamesmkrieger: Could you please tell me how to do that first.

SHZ66: from matplotlib.pyplot import gca

ax = gca()

prody/dynamics/plotting.py
SHZ66: Interesting. But I think the problem might be something else. I will look into this.

Did you use `ion` of `ioff`?

jamesmkrieger: Neither. I don't know what's the default.

jamesmkrieger: Thanks

prody/dynamics/plotting.py
jamesmkrieger: ok great. That works so I'll revert that change.

prody/dynamics/plotting.py
SHZ66: Try replacing `fig, ax = plt.subplots` to `fig = plt.figure`, that might fix it

SHZ66: Oh, forget to ask, in your code did you `show` everytime you make a subplot? Because if that's the case then multiple figures will be created. Calling `show` at the very end should fix the problem.

jamesmkrieger: Ah ok. That makes sense as long as auto_show is False. We are now doing that.

ajimoreno: ok we just tested it and that works perfectly

PR #585, opened 2018-03-22
updates to buildMSA

jamesmkrieger:
It can now use non-CA Atomic objects and MSA objects (it takes the first sequence). We probably need to tell the user with the LOGGER though or raise an error if they try a bigger MSA. I wasn't sure which was better so left it out.


prody/sequence/analysis.py
```diff
@@ -712,7 +712,7 @@ def alignSequencesByChain(PDBs, **kwargs):
     :type join_char: str
     """
     
-    if not isscalar(PDBs):
+    if not isinstance(PDBs, list) and not isinstance(PDBs, ndarray) and not isinstance(PDBs, tuple):
```
SHZ66: Is isscalar not working?
You can do this which is equivalent but shorter:
isinstance(PDBs, (list, ndarray, tuple))

prody/sequence/analysis.py
```diff
@@ -823,18 +823,28 @@ def buildMSA(sequences, title='Unknown', labels=None, **kwargs):
...
+                else:
+                    if len(sequence) > max_len:
+                        max_len = len(sequence)
 
             msa = []
             fetched_labels = []
             for i, sequence in enumerate(sequences):
                 if isinstance(sequence, Atomic):
-                    strseq = sequence.getSequence()
+                    strseq = sequence.ca.copy().getSequence()
```
SHZ66: sequence.ca should also have getSequence. copy will need to cost extra memory.

prody/sequence/analysis.py
```diff
@@ -908,7 +918,7 @@ def showAlignment(alignment, row_size=60, max_seqs=5, **kwargs):
 
     labels = kwargs.get('labels', None)
     if labels is not None:
-        if not isscalar(labels):
+        if not isinstance(labels, list) and not isinstance(labels, ndarray) and not isinstance(labels, tuple):
```
SHZ66: the same as above.

SHZ66:
Something like LOGGER.warn('msa contains more than one sequence; only the first is used') should be good. Raising an exception might be too strict I think.

prody/sequence/analysis.py
jamesmkrieger: Ok I’ll do that then. I tried it with a list instead of a tuple and that didn’t work

prody/sequence/analysis.py
jamesmkrieger: Ok true

prody/sequence/analysis.py
jamesmkrieger: Ok

jamesmkrieger:
Yes that LOGGER.warn option sounds good

prody/sequence/analysis.py
SHZ66: isscalar didn't work on a list? It should though. isscalar should be more comprehensive than manually checking enumerable types using isinstance. Do you have an example code that didn't work, and I will look into it. Because this is not the only place I use isscalar for type checking and if it is not working then I will replace them all.

prody/sequence/analysis.py
SHZ66: originally getSequence is exclusive to Chain type. But I have added it to all Atomic classes earlier so it should be safe.

prody/sequence/analysis.py
jamesmkrieger: Iterables are the opposite of scalars. Scalars are single entry things like int and str. If we use `if isscalar` instead of `if not isscalar`

SHZ66: You are absolutely correct. It is a mistake to use `not`. `isscalar` should work just fine.

PR #599, opened 2018-04-10
improvements to alignSequencesByChain

jamesmkrieger:


prody/dynamics/signature.py
```diff
@@ -1011,8 +1011,8 @@ def showSignatureVariances(mode_ensemble, **kwargs):
         sig = mode_ensemble.getVariances() 
     W = sig.getArray()[:, ::-1] # reversed to accommodate with matplotlib.pyplot.hist
     weights = np.ones_like(W)/float(len(W))
-
-    indices = mode_ensemble.getIndices()[0]
+    
+    indices = kwargs.pop('indices', mode_ensemble.getIndices()[0])
```
SHZ66: The way to define the indices of mode_ensemble is as such, for example if you want first three modes, then do:

``showSignatureVariances(mode_ensemble[:, :3])``

This is to be consistent with other ProDy functions, for example for mean square fluctuations calculations, you don't do something like ``calcSqFlucts(mode, indices=[1, 2, 3])``, instead you do ``calcSqFlucts(mode[:3])``.

PR #604, opened 2018-04-11
bug fix to n_modes='all'

jamesmkrieger:
If we use `lower()` with `is` then we get False. If we don't use `lower` then using `is` gives True.

### DIFF
--- a/prody/dynamics/anm.py
+++ b/prody/dynamics/anm.py
@@ -302,7 +302,7 @@ class ANMBase(NMA):
 
         if self._hessian is None:
             raise ValueError('Hessian matrix is not built or set')
-        if str(n_modes) is 'all':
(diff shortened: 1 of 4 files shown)


prody/dynamics/anm.py
```diff
@@ -302,7 +302,7 @@ def calcModes(self, n_modes=20, zeros=False, turbo=True):
 
         if self._hessian is None:
             raise ValueError('Hessian matrix is not built or set')
-        if str(n_modes) is 'all':
+        if str(n_modes).lower == 'all':
             n_modes = None
```
SHZ66: This one is missing brackets after lower. Yes, ‘is’ is bad for string comparison, especially when you have encode and binary strings to compare

jamesmkrieger: ok

PR #607, opened 2018-04-13
 added calcSignatureOverlaps and showSignatureOverlapsDiag 

jamesmkrieger:
Also tried to make the LOGGER show progress of matching modes but that doesn't seem to have worked.

### DIFF
--- a/prody/dynamics/signature.py
+++ b/prody/dynamics/signature.py
@@ -14,7 +14,7 @@ from .nma import NMA
 from .modeset import ModeSet
 from .mode import Mode, Vector
 from .functions import calcENM
-from .compare import calcSpectralOverlap, matchModes
+from .compare import calcSpectralOverlap, matchModes, calcOverlap
 
 from .analysis import calcSqFlucts, calcCrossCorr, calcFractVariance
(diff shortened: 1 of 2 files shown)


prody/dynamics/signature.py
```diff
@@ -901,8 +901,50 @@ def calcSignatureCrossCorr(mode_ensemble, norm=True):
...
+    n_modes = mode_ensemble.numModes()
+
+    overlaps = calcSignatureOverlaps(mode_ensemble)
+
+    meanV = np.diag(overlaps.mean().mean(axis=0))
+    stdV = np.diag(overlaps.std().std(axis=0))
+    maxV = np.diag(overlaps.max().max(axis=0))
+    minV = np.diag(overlaps.min().min(axis=0))
+
+    return _showSignature(meanV, stdV, minV, maxV)
```
SHZ66: I think you can just call ``showSignature`` here and ``_showSignature`` can remain to be a internal function of ``showSignature``. ``showSignature`` is designed to serve as what you are using ``_showSignature`` for.

SHZ66: Ok I see.

PR #608, opened 2018-04-13
added missed line in LOGGER progress for matching modes

jamesmkrieger:


jamesmkrieger:
Now I have the following:
@> all GNM modes were calculated for each of the 138 conformations in 10.96s.
@> Matching 359 modes across 138 modesets...
@> Matching 359 modes across 138 modesets... [ 28%] 271s

How do I get it to not repeat?

SHZ66:
Because there is also such report outside of the function in ModeEnsemble.match. I didn't put the messages in ``matchModes`` because the function is elementary so it could be called from everywhere, and a progress report might be too much.

SHZ66:
I feel like what you are doing with ``_showSignature`` can be done by ``showData`` actually.

jamesmkrieger:
Maybe. I don't know how though.

PR #614, opened 2018-04-17
Bug fixes and changed ModeEnsemble.reorder

SHZ66:
- Multiple bug fixes.
- ModeEnsemble.reorder now reorders based on eigenvalues.
- Temporarily turned off the tests for KDtree.

### DIFF
--- a/prody/apps/prody_apps/prody_pca.py
+++ b/prody/apps/prody_apps/prody_pca.py
@@ -4,6 +4,7 @@ graphical formats."""
 from ..apptools import *
 from .nmaoptions import *
 from . import nmaoptions
+from numbers import Integral
 
 try:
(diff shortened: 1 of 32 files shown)


SHZ66:
I think I may have figured out what the problem is. But for a complete fix, it requires a fix on biopython's side also, for which I submitted [a pull request](https://github.com/biopython/biopython/pull/1617)

SHZ66:
Ok, since it is impractical to wait for biopython to resolve the problems, I made their KDTree.c and KDTreemodule.c distributed with ProDy, and which we can control and fix by ourselves. Besides, based on setup.py and the license files, KDTree files should be redistributed with ProDy but somehow in some version we lost them.

PR #617, opened 2018-04-18
 updated fetchPfamPDBs to use current pdbmap.gz

jamesmkrieger:
Also renamed it to parsePfamPDBs

### DIFF
--- a/prody/database/pfam.py
+++ b/prody/database/pfam.py
@@ -501,13 +501,20 @@ def parsePfamPDBs(**kwargs):
 
     pdb_ids = [data_dict['PDB_ID'] for data_dict in data_dicts]
     chains = [data_dict['chains'] for data_dict in data_dicts]
-    result = parsePDB(*pdb_ids, chain=chains, header=True, **kwargs)
-                       .select('resnum {0} to {1}'
-                       .format(data_dict['PdbRange'].split('-')[0],
-                               data_dict['PdbRange'].split('-')[1])).copy()
+    ags, headers = parsePDB(*pdb_ids, chain=chains, header=True, **kwargs)
+
+    results = []
+    i = 0
+    for ag, header in zip(ags, headers):
+        results.append([])
+        results[i].append(ag.select('resnum {0} to {1}'
+                            .format(data_dicts[i]['PdbRange'].split('-')[0],
+                                    data_dicts[i]['PdbRange'].split('-')[1]))
+                            .copy())
+        results[i].append(header)
 
     if return_data:
-        return data_dict, result
+        return data_dict, results
     else:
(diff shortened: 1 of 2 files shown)


prody/database/pfam.py
```diff
@@ -473,36 +475,46 @@ def fetchPfamPDBs(**kwargs):
...
-    ftp.login('')
-    ftp.cwd('pub/databases/Pfam/mappings')
-    ftp.retrlines('RETR pdb_pfam_mapping.txt', data.append) 
+    ftp.login()
+    ftp.cwd('pub/databases/Pfam/current_release')
+    ftp.retrbinary('RETR pdbmap.gz', data_file.write)
+    ftp.quit()
+    data_file.close()
+
+    os.system('gunzip pdbmap.gz')
```
SHZ66: This line is platform specific. It’s better to use python gunzip.

prody/database/pfam.py
```diff
@@ -473,36 +475,46 @@ def fetchPfamPDBs(**kwargs):
                 raise ValueError('Please provide an integer for start or end when using query.')
 
     from ftplib import FTP
-    data = []
+    data_file = open('pdbmap.gz','wb')
```
SHZ66: This will generate a file on user’s working directory. If this file is intermediate, it is better to avoid to generate it. Please consider using file stream if there is no workaround.

prody/database/pfam.py
jamesmkrieger: I didn't find one but ok I'll have another look

prody/database/pfam.py
jamesmkrieger: I don't know how to use a stream from a gz file.

jamesmkrieger: It actually looks like there is no option to do that.

SHZ66:
Ok I will look at it later, but if this is the best way then it is okay.

jamesmkrieger:
I think it is

prody/database/pfam.py
```diff
@@ -473,36 +475,49 @@ def fetchPfamPDBs(**kwargs):
                 raise ValueError('Please provide an integer for start or end when using query.')
 
     from ftplib import FTP
-    data = []
+    import gzip
+
+    data_file = open('pdbmap.gz','wb')
```
SHZ66: I see the possibility of streaming the data here with [BytesIO](https://docs.python.org/release/3.1.3/library/io.html#binary-i-o), so everything can be done in memory and no need for memory and hard drive transmission. We can talk tomorrow on this.

jamesmkrieger: ok

Issue #553, opened 2018-03-07
Atom is the wrong type in extended anm


gsrule:
After extending normal modes to other atoms, e.g. CA to backbone, the new anm file is not usable because atoms has become a real (eg 836.0) and not an integer.  Thus it is not possible to calculate an ensemble for example.

gsrule:
Hi; I am following the following tutorial: http://prody.csb.pitt.edu/tutorials/enm_analysis/extend.html
(I substituted 1ubq for the pdb)
The initial (non-extended) works fine to make an ensemble:
In [9]: ensemble = sampleModes(anm[:3],p38_ca,n_confs=40,rmsd=0.8)
@> Parameter: rmsd = 0.80 A
@> Parameter: n_confs = 40
@> Modes are scaled by 1.3164091785390333.

However, if I extend it as follows:

In [10]: bb_anm,bb_atoms=extendModel(anm, p38_ca,p38.select('backbone'))

and then try to generate an ensemble, I get this:

In [13]: ensemble=sampleModes(bb_anm[:3],bb_atoms,n_confs=40,rmsd=0.8)
@> Parameter: rmsd = 0.80 A
@> Parameter: n_confs = 40
@> Modes are scaled by 2.839799250112639.
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-13-334642cc4447> in <module>()
----> 1 ensemble=sampleModes(bb_anm[:3],bb_atoms,n_confs=40,rmsd=0.8)

c:\python36\lib\site-packages\prody\dynamics\sampling.py in sampleModes(modes, atoms, n_confs, rmsd)
    372     if array.ndim > 1:
    373         for i in range(n_confs):
--> 374             append((array * scale * randn[i]).sum(1).reshape((n_atoms, 3)))
    375     else:
    376         for i in range(n_confs):

TypeError: 'float' object cannot be interpreted as an integer

SHZ66:
Thank you for the information. I was able to reproduce and track down the bug. It is fixed in #628. It will be incorporated in the next release. You can also do a quick fix on your local copy of ProDy.

SHZ66:
This problem should be fixed in the current [release](https://pypi.org/project/ProDy/#description).

PR #628, opened 2018-04-25
bug fix to NMA.setEigens

SHZ66:


SHZ66:
This fix relates to #553

PR #644, opened 2018-05-01
returned showPerturbResponseProfile from legacy to dynamics/plotting 

jamesmkrieger:
I think it is useful for PRS analysis. We use it for showing a row or column of the PRS matrix, which shows the effect of perturbing a particular residue on each of the others or the effect on a particular residue of perturbing each of the others, respectively.


SHZ66:
Actually it is integrated to `showPerturbResponse`. If you set `matrix=False` it will just plot the profile.

jamesmkrieger:
No, it doesn't show residue-specific profiles.

prody/atomic/functions.py
```diff
@@ -300,4 +303,58 @@ def sliceAtoms(atoms, select):
...
+    """Slice a matrix using indices extracted using sliceAtoms.
+
+    :arg matrix: any matrix (2D array)
+    :type matrix: `~numpy.ndarray`
+
+    :arg selection: a :class:`Selection` instance or selection string. 
+    :type selection: :class:`Selection`, str
+
+    :arg axis: the axis/direction you want to use to slice data from the matrix.
+        The options are 0 or 1 like in `~numpy` or 'both'. Default is 0 (row).
```
SHZ66: `~numpy` -> :mod:`~numpy`, also numpy uses `None` to mean "both", we may want to also make it consistent.

PR #663, opened 2018-05-09
 bug fix to showPerturbResponse 

jamesmkrieger:


prody/dynamics/plotting.py
```diff
@@ -1035,13 +1035,16 @@ def showPerturbResponse(model, atoms=None, show_matrix=True, select=None, **kwar
     else:
         if select is None:
             kwargs.pop('figure', 'effectiveness'); fig = gcf()
-            domain_bar = kwargs.pop('domain_bar', True)
+            domain = kwargs.pop('domain', False)
+            chain = kwargs.pop('chain', True)
```
SHZ66: I would recommend both default values to be `None` as it will be automatically determined by many factors.

Issue #461, opened 2017-12-07
How to get atomic charges from a PDB file


stephanmohr85:
The pdb format states that at the end of each line you can indicate an atomic charge:
'Columns 79 - 80 indicate any charge on the atom, e.g., 2+, 1-'
I was assuming that I could get these charge by using the command "getCharges()", but it doesn't work.
Using the following simple PDB file:

COMPND    free
AUTHOR    GENERATED BY OPEN BABEL 2.3.2
HETATM    1  C   LIG     1       0.000   0.000   0.000  1.00  0.00           C1+
MASTER        0    0    0    0    0    0    0    0    1    0    1    0
END

I get the following:

>>> import prody
>>> pdb = prody.parsePDB('C.pdb')
@> 1 atoms and 1 coordinate set(s) were parsed in 0.00s.
>>> print pdb.getCharges()
None

Is this a problem of ProDy or am I doing something wrong?

Any help is appreciated!

jamesmkrieger:
Hi,

This is a problem with ProDy. We had forgotten to include charges in the information obtained from PDB files because they are rarely there. This is now fixed in #675 and will be available in the next release.

Best wishes
James

jamesmkrieger:
Actually, this still doesn't work.

PR #675, opened 2018-05-14
added charges to _parsePDBLines when format is PDB

jamesmkrieger:
in response to #461

### DIFF
--- a/prody/proteins/pdbfile.py
+++ b/prody/proteins/pdbfile.py
@@ -529,8 +529,10 @@ def _parsePDBLines(atomgroup, lines, split, model, chain, subset,
                 hetero[acount] = startswith[0] == 'H'
                 segnames[acount] = line[72:76]
                 elements[acount] = line[76:78]
-                if line[78:80] is not '  ' and line[78:80] is not '':
+                try:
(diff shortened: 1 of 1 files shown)


prody/proteins/pdbfile.py
```diff
@@ -529,6 +529,10 @@ def _parsePDBLines(atomgroup, lines, split, model, chain, subset,
                 hetero[acount] = startswith[0] == 'H'
                 segnames[acount] = line[72:76]
                 elements[acount] = line[76:78]
+                try:
+                    charges[acount] = int(line[79] + line[78])
```
SHZ66: why is it 79 then plus 78?

jamesmkrieger: Because the sign is second in the PDB for some reason: http://www.wwpdb.org/documentation/file-format-content/format33/sect9.html#ATOM
"Columns 79 - 80 indicate any charge on the atom, e.g., 2+, 1-. In most cases, these are blank."

SHZ66: Ok sounds good. Just needed to make sure!

PR #718, opened 2018-06-05
 bug fix to parsePDB 

jamesmkrieger:
Without the fix it says it parsed 2 PDB files when it parses atomgroups and headers.


prody/proteins/pdbfile.py
```diff
@@ -138,8 +138,13 @@ def parsePDB(*pdb, **kwargs):
             results = results[0]
         results = list(results)
 
+        if len(results) == 2:
+            numPdbs = len(results[0])
+        else:
+            numPdbs = len(results)
+
```
SHZ66: This seems to be a little arbitrary: the user may have two pdbs without headers. In that case `len(results[0])` would probably obtain the number of atoms in the first structure (?). Could you determine it from the `header` argument or some other variables?

jamesmkrieger: I guess.

jamesmkrieger: How about `if isinstance(results[0][0], Atomic) and isinstance(results[1][0], dict):`

jamesmkrieger: ok that doesn't work if they are all headers

PR #766, opened 2018-10-24
added code for user data to Ensemble

jamesmkrieger:
I copied the data functions in Atomgroup and modified them to not have ATOMIC_FI and to have dimensions matching n_csets not n_atoms.


jamesmkrieger:
For some reason, my last commit doesn't work and the data isn't transferred over.

jamesmkrieger:
You were right it didn't work like that. Now it does.

SHZ66:
You could try saving this guy:
```
data = array([dict_data, 0])
```
This will trick numpy and let it believe it is an array already. Just look at how it stores `atoms` and you will see. The current version user data could overwrite other ensemble properties with the same name.

PR #791, opened 2019-01-02
bug fixes in getAngle and buildDistanceMatrix

jamesmkrieger:
Without this, you have to specify True or False to use getAngle. I choose default False because that's what getDihedral has. I also updated the documentation.

There are other commits here but they don't actually do anything.

### DIFF
--- a/prody/measure/measure.py
+++ b/prody/measure/measure.py
@@ -148,8 +148,8 @@ def calcAngle(atoms1, atoms2, atoms3, radian=False):
                     atoms3._getCoords(), radian)
 
 
-def getAngle(coords1, coords2, coords3, radian):
-    """Returns bond angle in degrees."""
+def getAngle(coords1, coords2, coords3, radian=False):
+    """Returns bond angle in degrees unless radian=True"""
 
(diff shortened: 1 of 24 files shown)


jamesmkrieger:
Not having a default value for radian arg in getAngle means that sometimes we get an error that we need three values and only have two. It's more a convenience thing than a proper bug I guess.

For buildDistanceMatrix, we had a problem that we get 3 rows instead of 1 when atoms2 is a single atom because the dimensions were getting taken from the lengths of atoms1 and atoms2.

prody/measure/measure.py
```diff
@@ -148,8 +152,8 @@ def calcAngle(atoms1, atoms2, atoms3, radian=False):
                     atoms3._getCoords(), radian)
 
 
-def getAngle(coords1, coords2, coords3, radian):
-    """Returns bond angle in degrees."""
+def getAngle(coords1, coords2, coords3, radian=False):
+    """Returns bond angle in degrees unless radian=True"""
```
SHZ66: Please put ``radian=True``. Same as below.

PR #799, opened 2019-01-28
bug fix to PDB header biomol parsing for non-Xray structures

jamesmkrieger:
The issue I had was that parsing multiple structures with biomol=True fails with an error if one of them has no biomol transformations in the header. I decided to take a conservative approach and still raise the error if it is an X-ray structure but only a warning if it's from another method. I don't know if this is the best solution or whether we should just make it a warning for all cases.

There may be other situations we should look at besides biomolecular transformations and work out a consistent philosophy for this. It might be better to handle errors differently e.g. parsePDB could receive a variable about them instead and decides what to do with it.

### DIFF
--- a/prody/proteins/header.py
+++ b/prody/proteins/header.py
@@ -1045,7 +1045,6 @@ def buildBiomolecules(header, atoms, biomol=None):
             LOGGER.warn("This is not an X-ray structure so no transformation was applied")
         else:
             raise ValueError("header doesn't contain biomolecular transformations")
-            return None
 
     if not isinstance(atoms, AtomGroup):
         atoms = atoms.copy()
--- a/prody/proteins/header.py
+++ b/prody/proteins/header.py
@@ -1041,10 +1041,8 @@ def buildBiomolecules(header, atoms, biomol=None):
 
     biomt = header.get('biomoltrans')
     if not isinstance(biomt, dict) or len(biomt) == 0:
-        if header.get('experiment') != 'X-RAY DIFFRACTION':
-            LOGGER.warn("This is not an X-ray structure so no transformation was applied")
-        else:
-            raise ValueError("header doesn't contain biomolecular transformations")
+        LOGGER.warn("no biomolecular transformations found so original structure was used")
+        return atoms
 
     if not isinstance(atoms, AtomGroup):
         atoms = atoms.copy()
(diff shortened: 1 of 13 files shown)


prody/proteins/header.py
```diff
@@ -1041,7 +1041,11 @@ def buildBiomolecules(header, atoms, biomol=None):
 
     biomt = header.get('biomoltrans')
     if not isinstance(biomt, dict) or len(biomt) == 0:
-        raise ValueError("header doesn't contain biomolecular transformations")
+        if header.get('experiment') != 'X-RAY DIFFRACTION':
+            LOGGER.warn("This is not an X-ray structure so no transformation was applied")
+        else:
+            raise ValueError("header doesn't contain biomolecular transformations")
+            return None
```
SHZ66: There is no need to return None if an error is already raised.

prody/proteins/header.py
```diff
@@ -1041,7 +1041,11 @@ def buildBiomolecules(header, atoms, biomol=None):
 
     biomt = header.get('biomoltrans')
     if not isinstance(biomt, dict) or len(biomt) == 0:
-        raise ValueError("header doesn't contain biomolecular transformations")
+        if header.get('experiment') != 'X-RAY DIFFRACTION':
```
SHZ66: Is this robust? Does every PDB file have an 'experiment' entry that indicates its type? And is the entry strictly spelt as 'experiment'?

prody/proteins/header.py
jamesmkrieger: Oh yeah

prody/proteins/header.py
jamesmkrieger: I think so

prody/proteins/header.py
jamesmkrieger: Now fixed

prody/proteins/header.py
jamesmkrieger: I've now replaced it with a warning always with returning of the original structure. We may want to return None instead. I'm not sure. It may also be something to handle in parsePDB when we provide multiple PDB IDs and only some of them have headers with biomoltrans. Perhaps returning None and then looking for None is a good way to go.

jamesmkrieger: I've now pushed here that change too.

jamesmkrieger:
This pull request is now only for changing the handling of biomol parsing and nothing else. What is does it gives a warning rather than raising an error so that it doesn't cause a problem for parsing multiple files.

PR #802, opened 2019-02-16
Development of go association database handling

jamesmkrieger:
There are still some problems but it largely works. 

The notebook showing usage, example results and problems is at https://pitt.box.com/s/t4gooijqhsotmpo3mtrt2ps3bpdkhxll. HTML output for easier visualising is at https://pitt.box.com/s/a0p9j06u8sf1ytbub7ete4c5u2v4pbo2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,6 @@ install:
   - conda create --yes -n test python=$TRAVIS_PYTHON_VERSION
   - source activate test
   - conda install --yes numpy scipy pip nose
-  - pip install goatools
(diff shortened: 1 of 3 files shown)


jamesmkrieger:
The checks fail because there's no module wget but I can pip install it on my system and that part works.

jamesmkrieger:
Ok, now the tests fail because it doesn't have the goatools module. How do we make it have it?

jamesmkrieger:
It now almost works. I'm just a bit confused that I'm reducing to a single number in a wrong place and not getting a matrix. See https://pitt.box.com/s/5cu6exah1p6vqwrl3s3k63qoo7mcrsk2

jamesmkrieger:
I was getting a single number because I was taking the mean at the end for some reason. Now it should be fine.

PR #803, opened 2019-02-16
A few minor bug fixes

jamesmkrieger:
I decided to separate my branches and this pull request from the commits about building biomolecular assemblies because we still need to check that what I did there is right.

### DIFF
--- a/prody/database/dali.py
+++ b/prody/database/dali.py
@@ -252,9 +252,12 @@ class DaliRecord(object):
         self._mapping = map_temp_dict
         self._data = data_list[3]
         lines = data_list[3].strip().split('\n')
+        # daliInfo = np.genfromtxt(lines[1:], delimiter = (4,3,6,5,5,5,6,5,57), usecols = [0,2,3,4,5,6,7,8], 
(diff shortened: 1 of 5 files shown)


jamesmkrieger:
My fix on showDomainBar isn't quite right. If no loc is specified then we have a problem of a half-unit shift.

jamesmkrieger:
I think everything is good here too. I can split it into fixes for separate issues if you prefer. I just found this way to do it much more cleanly and easily: https://gist.github.com/Loilo/930f141d9acf89e9e734ffa042acd750

SHZ66:
> I think everything is good here too. I can split it into fixes for separate issues if you prefer. I just found this way to do it much more cleanly and easily: https://gist.github.com/Loilo/930f141d9acf89e9e734ffa042acd750

Ok. I just think, to be safe, we probably should review the code sometime together. Nice useful link by the way - yes, in principle we should have separate pull requests for different problems - but I think this one is ok, no need to separate it.

jamesmkrieger:
ok

PR #810, opened 2019-03-18
Mass weighted calcGyradius bugs

mlevant:
Corrections to avoid:
1) ValueError (operands could not be broadcast together with shapes ...) when calling calcGyradius(p, weights=p.getMasses())
2) Uncorrect calculation of the (mass weighted) radius of gyration


SHZ66:
Thank you very much for catching this critical bug. I will merge it.

PR #828, opened 2019-04-17
Option to not renumber serials for writing PDB files

jamesmkrieger:
This could be useful for comparing PDB files containing selections back to full PDB files. Some programs use that too e.g. PLUMED for RMSD calculations: "Note that RMSD should be provided a reference structure in pdb format and can contain part of the system but the second column (the index) must reflect that of the full pdb so that PLUMED knows specifically which atom to drag where." (taken from https://plumed.github.io/doc-v2.5/user-doc/html/belfast-5.html#belfast-5-target)

### DIFF
--- a/prody/proteins/pdbfile.py
+++ b/prody/proteins/pdbfile.py
@@ -1182,7 +1182,7 @@ def writePDBStream(stream, atoms, csets=None, **kwargs):
         if multi:
             write('MODEL{0:9d}\n'.format(m+1))
         for i, xyz in enumerate(coords):
-            if i == 99999:
+            if pdbline != PDBLINE_GE100K and (i == 99999 or serials[i] >= 10000):
                 pdbline = PDBLINE_GE100K
             write(pdbline % (hetero[i], serials[i],
                          atomnames[i], altlocs[i],
--- a/prody/proteins/pdbfile.py
+++ b/prody/proteins/pdbfile.py
@@ -399,7 +399,6 @@ def _parsePDBLines(atomgroup, lines, split, model, chain, subset,
     if n_atoms > 0:
         asize = n_atoms
     else:
-        # most PDB files contain less than MAX_N_ATOM (99999) atoms
         asize = len(lines) - split
     addcoords = False
     if atomgroup.numCoordsets() > 0:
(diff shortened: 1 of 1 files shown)


prody/proteins/pdbfile.py
```diff
@@ -1172,9 +1182,9 @@ def writePDBStream(stream, atoms, csets=None, **kwargs):
         if multi:
             write('MODEL{0:9d}\n'.format(m+1))
         for i, xyz in enumerate(coords):
-            if i == 99999:
+            if pdbline != PDBLINE_GE100K and (i == 99999 or serials[i] >= 10000):
```
SHZ66: You missed a zero at the end.

prody/proteins/pdbfile.py
```diff
@@ -1172,9 +1182,9 @@ def writePDBStream(stream, atoms, csets=None, **kwargs):
...
                 pdbline = PDBLINE_GE100K
```
SHZ66: Raise a warning here saying the number of atoms is too big.

prody/proteins/pdbfile.py
jamesmkrieger: Thanks. This is now fixed.

prody/proteins/pdbfile.py
jamesmkrieger: This one likewise.

prody/proteins/pdbfile.py
```diff
@@ -1086,6 +1093,9 @@ def writePDBStream(stream, atoms, csets=None, **kwargs):
     if resnums is None:
         resnums = np.ones(n_atoms, int)
 
+    serials = atoms._getSerials()
+    if serials is None or renumber == True:
```
SHZ66: `if serials is None or renumber` is good. No need to compare `renumber` with `True`. It may give other developers the impression that `renumber` can be other value types.

jamesmkrieger:
I think actually it might make more sense to implement these changes on top of the hybrid36 changes later. Let's not merge this until we've discussed it.

SHZ66:
Ok. Although I tend to merge this one first, and then look at how we want to implement the hybrid36 feature.

prody/proteins/pdbfile.py
jamesmkrieger: done

PR #833, opened 2019-04-19
 bug fix on writePIR 

jamesmkrieger:
Otherwise it complains that types is not a list of strings


prody/sequence/msafile.py
```diff
@@ -764,7 +764,7 @@ def writePIR(filename, msa, **kwargs):
         raise TypeError('chain_sep should be a string or list of strings')
 
     types = kwargs.get('types', 'Sequence')
-    if types is None: 
+    if types is 'Sequence':
```
SHZ66: You need to use `types == 'Sequence'` to prevent potential bugs.

PR #838, opened 2019-04-29
bug fix for shifts in ANM and consistency fix in GNM

jamesmkrieger:
With shift 6, we get one mode too many.


SHZ66:
This won't change anything because `shift` will be updated later. And `shift` should equal to 6 because with the new fix, `shift` should be exactly the number of zero modes.

PR #839, opened 2019-04-29
Minor changes to parseEMD

jamesmkrieger:


prody/proteins/emdfile.py
```diff
@@ -48,12 +48,12 @@ def parseEMD(emd, **kwargs):
...
 
-    :arg return_map: Return the density map itself. Default is False in line with previous behaviour.
+    :arg map: Return the density map itself. Default is False in line with previous behaviour.
         This value is reset to True if make_nodes is False as something must be returned.
-    :type return_map: bool
+    :type map: bool
 
     :arg make_nodes: Use the topology representing network algorithm to fit pseudoatom nodes to the map.
-        Default is False and sets return_map to True.
+        Default is False and sets map to True.
```
SHZ66: True is a keyword and should be decorated by two stars on each side.

PR #849, opened 2019-05-08
added back hinge and show_hinge as options for showMode and showSqFlucts

jamesmkrieger:


jamesmkrieger:
We should maybe make this kind of change deeper in ProDy rather than popping the kwargs and changing individual functions but it's ok for now.

PR #851, opened 2019-05-09
changed default value of suppress_diag in showPerturbResponse

jamesmkrieger:


jamesmkrieger:
ok yep, that makes sense. Let's look into and talk about this further. Probably we want a different value depending on whether we're showing a matrix or a profile.

SnehaBheemireddy:
Screenshot from 2019-06-17 17-38-52

"select " option in PRS is not giving the expected result. I tried with the '3kg2' example given by you guys. I have attached the error message in the form of  a  picture here. please let me know the mistake.

jamesmkrieger:
Thanks for letting us know about this problem with the select option. This is not the right place to tell us about this though so I have made a new issue for it and will have a look into it further. This pull request is related to some other potential issue with showPerturbResponse.

SHZ66:
@SnehaBheemireddy Please look at #874 for more details about your problem.

PR #875, opened 2019-06-18
fixed writing to file for blastPDB in python 3

jamesmkrieger:


SHZ66:
Was this causing any error? I thought we could use blastpdb on Python 3 without any error.

PR #898, opened 2019-07-26
make sure on MacOSX C extensions are compiled using clang compilers

jamesmkrieger:
I just came across this commit in my old master. I don't know if it's helpful or not. We still need to find a way of checking what compilers are on the system and not just depend on system type.


SHZ66:
I added and removed these lines a while back. No, it doesn't help for our problem.

PR #908, opened 2019-08-06
added keyword to mapChainByChain to allow user-defined correspondence

jamesmkrieger:
Instead of finding chains with the same chain ID as the reference, we now have option of providing a set of chain IDs such that the first one provided is mapped to the first one in the reference, the second to the second and so on. The default behaviour is still the same.


jamesmkrieger:
I think this now works well.

PR #914, opened 2019-08-15
bug fix to matchModes 

jamesmkrieger:
so we can get indices without starting with modesets

Before, it would fail because ANM objects don't have getIndices. Now it can cope with that.

### DIFF
--- a/prody/dynamics/compare.py
+++ b/prody/dynamics/compare.py
@@ -375,7 +375,7 @@ def matchModes(*modesets, **kwargs):
     if index:
         try:
             ret = [modeset0.getIndices()]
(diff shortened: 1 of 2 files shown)


SHZ66:
Overall looks good to me, but I think we can either use a more specific try-except on AttributeError, or use hasattr to detect if the modeset has getIndices or not.

jamesmkrieger:
Like this?

SHZ66:
Yes. If this is bug-free then feel free to merge it.

jamesmkrieger:
ok, I'll check it a little more I guess

Issue #910, opened 2019-08-14
Fitting doesn't handle reflections properly


OrangeTowel:
Transformation.calcTransformation doesn't seem to handle reflections properly, see e.g. the following:

    obs_coords = np.array(
    [[-25.021704,  44.375935,  21.11747 ],
     [-24.512367,  44.058712,  19.75868 ],
     [-24.094004,  43.78701 ,  17.589916],
     [-22.865868,  43.461304,  18.257273],
     [-23.142508,  43.74437 ,  19.515787],
     [-25.816542,  44.013195,  15.985513],
     [-26.590445,  44.347313,  16.93461 ],
     [-26.443695,  44.324265,  18.255579],
     [-25.101994,  44.105095,  18.522114],
     [-24.495344,  43.723774,  16.260689],
     [-23.683393,  43.2716  ,  15.266071]])
    
    ideal_coords = np.array(
    [[-2.479,  5.346,  0.   ],
     [-1.912,  1.023,  0.   ],
     [-1.267,  3.124,  0.   ],
     [ 0.071,  2.771,  0.   ],
     [ 0.369,  1.398,  0.   ],
     [ 0.024,  4.897,  0.   ],
     [-0.668,  0.532,  0.   ],
     [-2.32 ,  2.29 ,  0.   ],
     [ 1.611,  0.909,  0.   ],
     [ 0.877,  3.902,  0.   ],
     [-1.291,  4.498,  0.   ]])

    T = pd.calcTransformation(ideal_coords, obs_coords)
    print(T.getRotation())

which prints
    
    [[ 0.90077067 -0.42857319  0.        ]
     [-0.0397204  -0.24241241  0.        ]
     [-0.43247485 -0.87037994  0.        ]]

which has determinant 0.
The problem can be traced down to file ``prody/measure/transform.py``, function ``getTransformation``, statement

    Id = np.array([[1, 0, 0],
                   [0, 1, 0],
                   [0, 0, np.sign(linalg.det(matrix))]])

which should in fact be 
    
    Id = np.array([[1, 0, 0],
                   [0, 1, 0],
                   [0, 0, np.sign(linalg.det(np.dot(np.transpose(Vh),U)))]])

See for example section 3.1.4 of DOI:10.1007/s001380050048 . The output with the corrected version above is
    
    [[ 0.90077067 -0.42857319 -0.07026543]
     [-0.0397204  -0.24241241  0.96935985]
     [-0.43247485 -0.87037994 -0.23538109]]

[...]

PR #915, opened 2019-08-16
fix to calcTransformation to correctly handle reflection

jamesmkrieger:
This fix was suggested in issue #910


SHZ66:
The title should be "correctly" handle reflection. I will look at this.

jamesmkrieger:
ok I changed the title

SHZ66:
Yes, this is indeed a valid fix to the implementation of the Kabsch algorithm.

SHZ66:
The only problem might be that `Vh` is already transposed right eigenvector matrix so it doesn't need to be transposed again.

Issue #916, opened 2019-08-16
showMatrix with y_array as tree doesn't work well enough


jamesmkrieger:
I get the following error:

```
TypeError                                 Traceback (most recent call last)
<ipython-input-1-e92eb9b2363c> in <module>()
     31
     32 plt.figure()
---> 33 showMatrix(expt_rmsd_matrix, y_array=expt_rmsd_tree)
     34 labels = [label[:4] for label in expt_ens.getLabels()]
     35 plt.xticks(range(expt_ens.numConfs()), labels)

/home/jkrieger/Documents/code/ProDy/prody/utilities/catchall.pyc in showMatrix(matrix, x_array, y_array, **kwargs)
    407
    408             y = y_array
--> 409             xp, yp = interpY(y)
    410             points = np.array([yp, xp]).T.reshape(-1, 1, 2)
    411             segments = np.concatenate([points[:-1], points[1:]], axis=1)

/home/jkrieger/Documents/code/ProDy/prody/utilities/misctools.pyc in interpY(Y)
    451
    452 def interpY(Y):
--> 453     Y = asarray(Y, dtype=float)
    454     n = len(Y)
    455     X = arange(n)

/home/jkrieger/anaconda2/lib/python2.7/site-packages/numpy/core/numeric.pyc in asarray(a, dtype, order)
    499
    500     """
--> 501     return array(a, dtype, copy=False, order=order)
    502
    503

TypeError: float() argument must be a string or a number
```

I think this is because the check for checking whether to check for trees is wrong. On line 328, the `np.isscalar` check fails because a tree isn't a scalar. We might want to use `hasattr` instead.

We also don't really need to import `Phylo` as we have a function `showTree` that's based on `Phylo.draw`.

jamesmkrieger:
To clarify, we end up on line 409 because tree_mode is False.

jamesmkrieger:
We could also use `np.size` instead of `np.isscalar` as a tree has size 1.

SHZ66:
A type checker might be more suitable in this case.

jamesmkrieger:
Yes, maybe we should have our own type checker for trees that doesn't depend on Biopython.

jamesmkrieger:
Using `np.size` does work although now I have another error:

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-1-0ba29fd14ef2> in <module>()
     31
     32 plt.figure()
[...]

PR #917, opened 2019-08-16
temporary fix for showMatrix with tree

jamesmkrieger:


jamesmkrieger:
This now fixes most of the problems from issue #916 and can be merged. 
working showMatrix with tree

We, of course, need to remember to use `origin='upper'` so the matrix is oriented the same way as the tree.

PR #918, opened 2019-08-21
Adaptive ANM

jamesmkrieger:
This now works pretty well and I think it's probably ready for more users. Documentation is still lacking in many places but otherwise it's good.


SHZ66:
I am merging this one since it is a separate module.

PR #919, opened 2019-08-23
added function trimCombineEnsembles

jamesmkrieger:
This function uses mapping and trimming to combine ensembles. 

I also reorganised the documentation on some of the other ensemble functions. We still need to add documentation for mapping_func, which is a positional argument in some cases and a kwarg in others. I don't know what we do for type seeing as it would take functions as input.


jamesmkrieger:
This function doesn't quite work. I still need to fix some things but it's almost there.

jamesmkrieger:
Actually it still doesn't quite work.

PR #931, opened 2019-10-06
fix typing in pdbclusters

JoshuaMeyers:
# Bug while attempting to retrieve clusters from pdb
```
In [1]: from prody.proteins import pdbclusters

In [2]: pdbclusters.listPDBCluster('1atp', 'I', 30)
AssertionError                            Traceback (most recent call last)
<ipython-input-2-5cd1adf066f4> in <module>
----> 1 pdbclusters.listPDBCluster('1atp', 'I', 30)

/miniconda3/lib/python3.7/site-packages/prody/proteins/pdbclusters.py in listPDBCluster(pdb, ch, sqid)
     80     clusters = PDB_CLUSTERS[sqid]
     81     if clusters is None:
---> 82         loadPDBClusters(sqid)
     83         clusters = PDB_CLUSTERS[sqid]
     84     pdb_ch = pdb.upper() + '_' + ch.upper()

/miniconda3/lib/python3.7/site-packages/prody/proteins/pdbclusters.py in loadPDBClusters(sqid)
     26         LOGGER.info('Loading all PDB sequence clusters.')
     27     else:
---> 28         assert isinstance(sqid, int), 'sqid must be an integer'
     29         if sqid not in PDB_CLUSTERS:
     30             raise ValueError('PDB cluster data is not available for sequence '

AssertionError: sqid must be an integer
```
Two typing issues are solved by this PR:
1. The sequence identity variable `sqid` is converted to a `numpy.int64` type by `listPDBCluster` when it is required to be an `int` (by `loadPDBClusters`).
2. The second stage of this function performs substring searches on a bytestring.  The queries are not bytestrings so this causes a second set of type errors.

This PR converts the numpy integer to `int` and all substring queries to bytestring objects

_user spec:
python version: 3.7
prody version: 1.10.10
os: mac os x_

### DIFF
--- a/prody/proteins/pdbclusters.py
+++ b/prody/proteins/pdbclusters.py
@@ -5,7 +5,7 @@ import os.path
 
 from numpy import array, abs
 
-from prody import LOGGER, SETTINGS, getPackagePath
+from prody import LOGGER, getPackagePath
 from prody.utilities import openFile, openURL
 
 __all__ = ['fetchPDBClusters', 'loadPDBClusters', 'listPDBCluster']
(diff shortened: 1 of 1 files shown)


SHZ66:
Thanks a lot for the fix. Regarding the first one, I think it makes more sense to make the type checker in `loadPDBClusers` more general. I will make that change and accept this one too.

SHZ66:
I am closing this PR for now but let me know if there is any problem.

JoshuaMeyers:
Yep seems to work now, cheers

Issue #924, opened 2019-09-20
writePSF does not write angles or dihedrals


alexsaadfalcon:
Really throws a wrench in things when trying to manipulate a PDB and PSF and have the results be compatible. I have to edit the PDB in ProDy and then use vmd scripting to edit the PSF, so please fix this.

jamesmkrieger:
ok. We'll look into this soon.

jamesmkrieger:
It seems like we never implemented angles and dihedrals into Atomic objects. I will implement these soon. Sorry for the inconvenience.

jamesmkrieger:
I have started a pull request #933, which includes a start towards this. I will add more to it soon and let you know when I am done.

jamesmkrieger:
Thanks for reminding us about this.

jamesmkrieger:
I believe we are there now.

jamesmkrieger:
We have just merged the pull request introducing these features. Please can you test it and let us know if you have any issues.

jamesmkrieger:
We have another pull request (#1034), which fixes some more issues with this. I think we're basically there now but it would be good if you could test and confirm, please.

PR #933, opened 2019-10-07
added handling of bonds, angles, dihedrals and impropers

jamesmkrieger:


jamesmkrieger:
I am starting to make progress towards addressing  #924. We still need the following:

1. Addition of angles, dihedrals and impropers to writePSF

2. Handling of angles, dihedrals and impropers in selection algebra

jamesmkrieger:
4. Update saveEnsemble and maybe other save and write functions to handle this data

jamesmkrieger:
My last commit includes reading and writing angles, dihedrals and impropers actually so I think we're done.

jamesmkrieger:
Oh actually not quite

jamesmkrieger:
We could still do with other terms. The ones from autopsf have the following:

       0 !NDON: donors
       0 !NACC: acceptors
       0 !NNB
       1       0 !NGRP
      244 !NCRTERM: cross-terms

The last one is documented on https://www.ks.uiuc.edu/Training/Tutorials/namd/namd-tutorial-unix-html/node23.html but the others I'm not sure about. I'm guess donors and acceptors are single atom properties.

jamesmkrieger:
This one describes donors, acceptors and non-bonded exclusions too: http://www.charmm-gui.org/?doc=lecture&module=pdb&lesson=6

jamesmkrieger:
I'm happy to call this done now. There are still other features of a PSF file that we could add in theory but it's probably ok.

jamesmkrieger:
Actually my bug fix isn't quite there.

jamesmkrieger:
This works now!

jamesmkrieger:
We still have more work to do here actually. I just got this error that I'm going to address next.

```
~/Documents/code/ProDy/prody/trajectory/psffile.py in writePSF(filename, atoms)
    322             write('\n')
    323 
--> 324     angles = list(atoms._iterAngles())
    325     if angles:
    326         angles = array(angles, int) + 1

~/Documents/code/ProDy/prody/atomic/atomic.py in __getattribute__(self, name)
    127                              'is not a valid selection string'
    128                              .format(self.__class__.__name__, name,
--> 129                                      repr(selstr)))
    130 
    131     def __getstate__(self):

AttributeError: Selection object has no attribute `_iterAngles` and '_iterAngles' is not a valid selection string
```

jamesmkrieger:
The way to do this is to make such functions in atompointer, which is the superclass for atom, atomsubset and selection based on its existing _iterBonds and iterBonds functions.

jamesmkrieger:
Probably good to go now.

PR #935, opened 2019-10-08
bug fix to showAlignment

jamesmkrieger:


jamesmkrieger:
There's something else that needs fixing too as it still doesn't behave.

PR #941, opened 2019-10-20
Get masses from IUPACData

mlevant:
getMasses() is modified to return the atomic mass of any element using the 'atom_weights' dictionary from Bio.Data.IUPACData
( currently approximate masses are returned for most comment elements only: C, N, S, O, H )


jamesmkrieger:
Thanks for doing this.

SHZ66:
Thanks for the modification. I am accepting it as is.

PR #946, opened 2019-10-25
Use correct operator to compare strings

cornett:


SHZ66:
Thanks a lot for the fix. I am accepting this.

PR #954, opened 2019-11-12
set default behaviour of _parseEMDLines based on n_nodes

jamesmkrieger:
If no kwargs are provided to parseEMD then the default values for _parseEMDLines mean that we return the raw EMD map but if the user sets n_nodes > 0 then it make nodes with TRN and does not return the EMD map unless the user specifically ask for it.


jamesmkrieger:
This behaviour should be in line with the website.

Issue #798, opened 2019-01-27
Inferring bonds in AtomGroups


npabon:
Is there a way to infer covalent bonds from inter-atomic distances in an AtomGroup? Alternatively, can Prody parse the CONECT records in a PDB file?

I'm looking for an alternative to AtomGroup.setBonds() that does not require me to create a list of atom index pairs.

SHZ66:
Hi Nick,

I don't think there is a function in ProDy to do that. Neither does ProDy parse the CONECT records. I am adding this as a feature request.

jamesmkrieger:
Hi Nick,

ProDy does have a bonds module and can parse bond information from PSF files. What you could do is use the AutoPSF module in VMD to generate a PSF file and then use that as input for ProDy.

Inferring bonds would require knowledge about atom types and topologies not just distances, and would be quite extensive to implement in ProDy most likely.

Best wishes
James

SHZ66:
In #961, `parsePDB` is now able to parse the `CONECT` entries. However, we still need to figure out how to infer the covalent bonds within and among standard residues. We could use either bond length or infer from the amino acid types.

jamesmkrieger:
We could maybe use charmm36 topology files. I guess we could also have a wrapper to AutoPSF or OpenBABEL or something.

jamesmkrieger:
I have actually have a solution to this from my MolSSI fellowship bootcamp: https://github.com/jamesmkrieger/MoleCool/blob/master/molecool/molecule.py

PR #961, opened 2019-11-21
Enabled parsePDB to evaluate CONECT entries

SHZ66:
Working towards solving #798.

- Now `parsePDB(..., bonds=True)` can evaluate `CONECT` entries in the PDB files. These entries are interpreted as the `Bond` objects in ProDy.
- Added a setting called `auto_binds`, which when set to `True`, `parsePDB` will read `CONECT` entries by default.

### DIFF
--- a/prody/__init__.py
+++ b/prody/__init__.py
@@ -144,6 +144,7 @@ CONFIGURATION = {
     'typo_warnings': (True, None, None),
     'check_updates': (0, None, None),
     'auto_secondary': (False, None, None),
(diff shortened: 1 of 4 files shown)


dkoes:
```
pdb = prody.parsePDB('4iuf',bonds=True)
list(pdb.iterBonds())
```
This PDB has CONECT records, but this resulting list above is empty.  Also, it would be nice if we could iterBonds on atom selections.

PR #971, opened 2019-12-03

jamesmkrieger:
It already has a return higher up for just slicing the covariance.


SHZ66:
I am closing this one since we've discussed this block is still useful.

PR #977, opened 2019-12-20
added hierarchical clustering functionality to showMatrix

oacar:
needed to show PRS heatmap with hierarchical clustering dendrograms, so I edited showMatrix function.  if both x_array and y_array are given, user can ask row or column clustering using cluster_row/cluster_col kwargs. generates following output. 

Screen Shot 2019-12-20 at 12 06 03 PM

### DIFF
--- a/prody/dynamics/plotting.py
+++ b/prody/dynamics/plotting.py
@@ -1119,7 +1119,9 @@ def showPerturbResponse(model, atoms=None, show_matrix=True, select=None, **kwar
         show = showAtomicMatrix(prs_matrix, x_array=sensitivity, 
                                 y_array=effectiveness, atoms=atoms, 
                                 **kwargs)
-        xlabel('Residues')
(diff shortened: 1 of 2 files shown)


jamesmkrieger:
This functionality sounds great. Thanks for adding it. I had some other ways for doing similar things but I think this could be better.

It says you have conflicts that need to be resolved. Please could you fix those?

Many thanks
James

SHZ66:
> This functionality sounds great. Thanks for adding it. I had some other ways for doing similar things but I think this could be better.
> 
> It says you have conflicts that need to be resolved. Please could you fix those?
> 
> Many thanks
> James

Yes, we have something similar in `showMatrix` to display a tree on the x- and y-axis and the `calcTree` to do the clustering. But this is also well implemented and @oacar has kindly solved the conflicts, so I am accepting it as is. 

In the long run, we may want to keep calculating and showing functions separated. We may want to add hierarchical clustering to the `calcTree` function.

PR #980, opened 2020-01-23

jamesmkrieger:
There is another if condition for PCA higher up so this one is never used.


SHZ66:
I am closing this one for now as per discussion which we decided to leave this as is. reduceModel on PCA isn't very useful but at least it should make the dimension consistent with applying it on other models.

PR #990, opened 2020-03-26

jamesmkrieger:
Somehow this is still in my devel-10 branch so I guess it still needs doing.


SHZ66:
I am closing this one for now with the same reason as #971. As per discussion which we decided to leave this as is. reduceModel on PCA isn't very useful but at least it should make the dimension consistent with applying it on other models. Plus this pull seems to contain a lot commits that have been merged into main prody already. So merging it might not be safe.

PR #1020, opened 2020-05-04
psi-blast changes for PY3K

jamesmkrieger:
We may need to test these more but it seems like they fixed the problem I was having. In any case, psi-blast isn't really working properly so this will be a continuing thing.


SHZ66:
Is there a reason why changing the default behavior of filter from 'F' to 'no'?

PR #1023, opened 2020-05-05
added adjust_text to showProjection and showCrossProjection

jamesmkrieger:
This module adjusts text so that it doesn't overlap and changes with zooming. The output becomes 
pca_proj2d_adjust instead of 

pca_proj2d


jamesmkrieger:
I just checked and it works for 3D projections too

PR #1024, opened 2020-05-05
added xtickrotation to showMatrix

jamesmkrieger:
Allows us to get
rmsd_matrix_xtickrotation
instead of
rmsd_matrix

### DIFF
--- a/prody/utilities/catchall.py
+++ b/prody/utilities/catchall.py
@@ -486,6 +486,10 @@ def showMatrix(matrix, x_array=None, y_array=None, **kwargs):
 
     :arg interactive: turn on or off the interactive options
     :type interactive: bool
+
+    :arg xtickrotation: turn on or off rotation of the xticklabels
+                        default is False
+    :type xtickrotation: bool
     """
 
     from matplotlib import ticker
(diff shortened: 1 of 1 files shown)


jamesmkrieger:
The only way I found to alleviate the shift was to use 
`plt.savefig('rmsd_matrix', bbox_inches='tight')`

Ideally, there should be something to fix it on the figure itself. Without that we get the following.
rmsd_matrix_xtickrotation_no_shift

SHZ66:
> The only way I found to alleviate the shift was to use
> `plt.savefig('rmsd_matrix', bbox_inches='tight')`
> 
> Ideally, there should be something to fix it on the figure itself. Without that we get the following.
> rmsd_matrix_xtickrotation_no_shift

The visuals of plots are subject to the layouts and sizes of figures. It may look good on a particular setup but bad on others. Sometimes it is just impossible to make it look good on all occasions. So I don't think we should worry too much about how the plots appear but rather provide flexibility for the users to manipulate them (that is why we expose the matplotlib objects to them so they can play it with themselves).

jamesmkrieger:
Makes sense

PR #1027, opened 2020-05-05
bug fix for getModeSets with int index

jamesmkrieger:
The error otherwise is

```
In [64]: anm_ens.getModeSets(0)                                                                                    
AttributeError                            Traceback (most recent call last)
<ipython-input-64-78d2e854bb26> in <module>
----> 1 anm_ens.getModeSets(0)

~/Documents/code/ProDy/prody/dynamics/signature.py in getModeSets(self, index)
    229         if index is None:
    230             return self._modesets
--> 231         return self[index]._modesets
    232 
    233     def getArray(self, mode_index=0):

AttributeError: 'ModeSet' object has no attribute '_modesets'
```

### DIFF
--- a/prody/dynamics/signature.py
+++ b/prody/dynamics/signature.py
@@ -228,9 +228,12 @@ class ModeEnsemble(object):
 
         if index is None:
             return self._modesets
-        elif isinstance(index, Integral):
-            return self[index]
-        return self[index]._modesets
+
+        subset = self[index]
+        if isinstance(subset, ModeSet):
+            return subset
+        else:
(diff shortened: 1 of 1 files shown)


jamesmkrieger:
Oh I also cleaned up and removed an unnecessary variable.

SHZ66:
please use `isinstance(index, numbers.Integral)` instead. Or 

```
subset = self[index]
if isinstance(subset, ModeSet):
    return subset
else:
    ...
```

is probably better. Because there may be other ways to slice the object into a single element beside using an int index.

PR #1033, opened 2020-05-11
Some fixes to emdfile

jamesmkrieger:


jamesmkrieger:
Actually I may want to split these up and check them. I've made other fixes to this and these may not all be compatible.

jamesmkrieger:
I think it should be good now

PR #1034, opened 2020-05-11
bug fixes on pointer PSF terms and PSF reading and writing

jamesmkrieger:


SHZ66:
`if bonds != [None]:` is not a good way of checking the value. Is there a way to avoid this? For example, can bonds be an empty list or just None when absent?

jamesmkrieger:
They now become an empty list.

PR #1035, opened 2020-05-12
added diag kwarg option to showSignatureOverlaps

jamesmkrieger:
That way we can have a matrix.

### DIFF
--- a/prody/dynamics/compare.py
+++ b/prody/dynamics/compare.py
@@ -58,12 +58,12 @@ def calcOverlap(rows, cols, diag=False):
         cols = cols.getArray()
     cols *= 1 / (cols ** 2).sum(0) ** 0.5
 
-    overlap_matrix = np.dot(rows.T, cols)
-
     if diag:
-        return np.diag(overlap_matrix)
+        overlaps = np.einsum('ij,ji->i', rows.T, cols)
+    else:
+        overlaps = np.dot(rows.T, cols)
     
-    return overlap_matrix
+    return overlaps
 
 
 def printOverlapTable(rows, cols):
--- a/prody/dynamics/compare.py
+++ b/prody/dynamics/compare.py
@@ -20,11 +20,14 @@ __all__ = ['calcOverlap', 'calcCumulOverlap', 'calcSubspaceOverlap', 'calcSpectr
 SO_CACHE = {}
 WO_CACHE = {}
 
-def calcOverlap(rows, cols):
+def calcOverlap(rows, cols, diag=False):
     """Returns overlap (or correlation) between two sets of modes (*rows* and
     *cols*).  Returns a matrix whose rows correspond to modes passed as *rows*
     argument, and columns correspond to those passed as *cols* argument.
-    Both rows and columns are normalized prior to calculating overlap."""
+    Both rows and columns are normalized prior to calculating overlap.
+    
+    This function can now return the diagonal of the overlap matrix if *diag*
+    is set to **True**."""
 
     if not isinstance(rows, (NMA, ModeSet, Mode, Vector, np.ndarray)):
         raise TypeError('rows must be NMA, ModeSet, Mode, Vector, or array, not {0}'
@@ -55,7 +58,12 @@ def calcOverlap(rows, cols):
         cols = cols.getArray()
     cols *= 1 / (cols ** 2).sum(0) ** 0.5
 
-    return np.dot(rows.T, cols)
+    overlap_matrix = np.dot(rows.T, cols)
+
+    if diag:
+        return np.diag(overlap_matrix)
+    
+    return overlap_matrix
 
 
 def printOverlapTable(rows, cols):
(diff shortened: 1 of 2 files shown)


SHZ66:
If we are going to have this option, we should consider just calculate the diagonal to have some speed gain.

jamesmkrieger:
The original behaviour was to have diag being `True` though.

jamesmkrieger:
Previously, we would only use this function on matched modes to generate the curves like in the SignDy paper. The new behaviour, in this case showing the whole overlap table matrix, usually shouldn't be the default one. We also want the default of `showSignatureOverlap` to match the default of `calcSignatureOverlap`, which also has `diag=True`.

jamesmkrieger:
Yes, definitely we should just calculate the diagonal. `calcSignatureOverlap` already has the `diag` option but doesn't actually do that. I'll see what I can do about it.

jamesmkrieger:
You're right it makes a huge difference.

New implementation:
```
Python 3.7.1 (default, Dec 14 2018, 13:28:58) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.13.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import os 
   ...: import time 
   ...: from prody import * 
   ...: import matplotlib.pyplot as plt 
   ...: import numpy as np 
   ...: plt.ion() 
   ...:  
   ...: proc = 5 
   ...: a3_dir = './'                                                                           

In [2]: mode_ensemble = anm_ens_20 = loadModeEnsemble(a3_dir+'A3_dimers_anm_20.modeens.npz')    

In [3]: start = time.time()  
   ...: for i in range(10): 
   ...:     overlap_mat = calcSignatureOverlaps(mode_ensemble, diag=False) 
   ...: end = time.time(); duration = end - start; print(duration/10)                           
2.26108660697937

In [4]: start = time.time()  
   ...: for i in range(10): 
   ...:     overlap_diag = calcSignatureOverlaps(mode_ensemble, diag=True) 
   ...: end = time.time(); duration = end - start; print(duration/10)                           
2.391601014137268
```

Old implementation:
```
Python 3.7.1 (default, Dec 14 2018, 13:28:58) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.13.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import os 
   ...: import time 
   ...: from prody import * 
   ...: import matplotlib.pyplot as plt 
   ...: import numpy as np 
   ...: plt.ion() 
   ...:  
   ...: proc = 5 
   ...: a3_dir = './'                                                                           

In [2]: mode_ensemble = anm_ens_20 = loadModeEnsemble(a3_dir+'A3_dimers_anm_20.modeens.npz')    

In [3]: start = time.time()  
   ...: for i in range(10): 
   ...:     overlap_mat = calcSignatureOverlaps(mode_ensemble, diag=False) 
   ...: end = time.time(); duration = end - start; print(duration/10)                           
2.2047492980957033

In [4]: start = time.time()  
   ...: for i in range(10): 
   ...:     overlap_diag = calcSignatureOverlaps(mode_ensemble, diag=True) 
   ...: end = time.time(); duration = end - start; print(duration/10)                           
46.54589171409607
```

jamesmkrieger:
This is the timings for the einsum version. NB: the second set of numbers is for the full 2244 modes.

```
Python 3.7.1 (default, Dec 14 2018, 13:28:58) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.13.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import os 
   ...: import time 
   ...: from prody import * 
   ...: import matplotlib.pyplot as plt 
   ...: import numpy as np 
   ...: plt.ion() 
   ...:  
   ...: proc = 5 
   ...: a3_dir = './'                                                                                                             

In [2]: mode_ensemble = anm_ens_20 = loadModeEnsemble(a3_dir+'A3_dimers_anm_20.modeens.npz')                                      

In [3]: start = time.time()  
   ...: for i in range(10): 
   ...:     overlap_mat = calcSignatureOverlaps(mode_ensemble, diag=False) 
   ...: end = time.time(); duration = end - start; print(duration/10)                                                             
2.4902858972549438

In [4]: start = time.time()  
   ...: for i in range(10): 
   ...:     overlap_diag = calcSignatureOverlaps(mode_ensemble, diag=True) 
   ...: end = time.time(); duration = end - start; print(duration/10)                                                             
2.361295294761658

In [5]: mode_ensemble = anm_ens = loadModeEnsemble(a3_dir+'A3_dimers_anm.modeens.npz')                                            

In [6]: start = time.time()  
   ...: for i in range(10): 
   ...:     overlap_mat = calcSignatureOverlaps(mode_ensemble, diag=False) 
   ...: end = time.time(); duration = end - start; print(duration/10)                                                             
18.004747581481933

In [7]: start = time.time()  
   ...: for i in range(10): 
   ...:     overlap_diag = calcSignatureOverlaps(mode_ensemble, diag=True) 
   ...: end = time.time(); duration = end - start; print(duration/10)                                                             
8.800933694839477
```

PR #1037, opened 2020-05-13
bug fixes to showProjection around colors

jamesmkrieger:
An update to matplotlib means cmap changed and it doesn't have an attribute `.color`. This is the fix to the problem. It wasn't necessary to change anything else.

### DIFF
--- a/prody/dynamics/plotting.py
+++ b/prody/dynamics/plotting.py
@@ -7,7 +7,7 @@ and are prefixed with ``show``.  Function documentations refers to the
 and keyword arguments are passed to the Matplotlib functions."""
 
 from collections import defaultdict
-
+from numbers import Integral
 import numpy as np
 
 from prody import LOGGER, SETTINGS, PY3K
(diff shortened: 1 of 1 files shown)


jamesmkrieger:
This still doesn't work.

jamesmkrieger:
This now works in 2D and 3D:
time-colored_run1_pca_proj3d
time-colored_run1_pca_proj2d

It raises lots of these warnings in 2D though:
```'c' argument looks like a single numeric RGB or RGBA sequence, which should be avoided as value-mapping will have precedence in case its length matches with 'x' & 'y'.  Please use a 2-D array with a single row if you really want to specify the same RGB or RGBA value for all points.```

**Comment by jamesmkrieger** [comment]
The 2D case gives lots of these warnings but otherwise everything is good.
```'c' argument looks like a single numeric RGB or RGBA sequence, which should be avoided as value-mapping will have precedence in case its length matches with 'x' & 'y'.  Please use a 2-D array with a single row if you really want to specify the same RGB or RGBA value for all points.```

**Comment by SHZ66** [comment]
```
from numbers import Number

isinstance(3.0, Number)
```

SHZ66:
[matplotlib.pyplot.scatter](https://matplotlib.org/3.2.0/api/_as_gen/matplotlib.pyplot.scatter.html)

Issue #1042, opened 2020-05-17
calcPerturbResponse on extrenally supplied covariance matrix


mmagithub:
Hi,
I am trying to use the calcPerturbResponse function on a covariance matrix generated from an MD simulation using "WORDOM". The matrix shape is (3642, 3642).
I tried to use the function directly on this matrix, but it keeps complianing that an ANM or GNM instances have to exist. 

I tried:
covar_mat = np.genfromtxt('pca1-matrix.txt')
calcPerturbResponse(covar_mat)

But I am getting:

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-83-92fb1a68b645> in <module>
      1 covar_mat = np.genfromtxt('pca1-matrix.txt')
----> 2 calcPerturbResponse(covar_mat)

~/anaconda3/envs/my-rdkit-env/lib/python3.8/site-packages/prody/dynamics/perturb.py in calcPerturbResponse(model, atoms, **kwargs)
     47 
     48     if not isinstance(model, (NMA, ModeSet, Mode)):
---> 49         raise TypeError('model must be an NMA, ModeSet, or Mode instance')
     50 
     51     if isinstance(model, NMA) and len(model) == 0:

TypeError: model must be an NMA, ModeSet, or Mode instance

Any clue how can I do that.

Thanks,
marawan

jamesmkrieger:
Hi Marawan,

I think I need to change the code to let you do that. I'll have a look and get back to you.

Best wishes
James

jamesmkrieger:
I think my changes should work now so this functionality should be available to you soon. Thanks for your pull request.

jamesmkrieger:
Actually, the best way to do this is to provide the matrix to a PCA model. I've updated the docs in #1044 to explain this.

jamesmkrieger:
(#1043 where I updated the code is closed without merging as we can do this without that)

jamesmkrieger:
We are also planning to update the online tutorial to help with this.

mmagithub:
Thanks James, I tried the following:

pca = PCA()
pca.setCovariance(covar_matrix)
pca.calcModes()
prs_mat, eff, sens = calcPerturbResponse(pca)

But I got this error:

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-39-09c483a4add3> in <module>
      2 pca.setCovariance(covar_matrix)
      3 pca.calcModes()
[...]

PR #1046, opened 2020-05-21
made PCA setCovariance set _n_atoms as int

jamesmkrieger:
This is a problem we realised because of #1042


jamesmkrieger:
Do we want to do this or use `//` as in anm.py? It seems to me that floor division with `//` is a bad idea so we should change that one too.

SHZ66:
I would say floor division is always preferred, because int involves data type conversion which should be avoided.

PR #1056, opened 2020-06-04
added superpose False option to buildPDBEnsemble

jamesmkrieger:
That way, the user has the option to not superpose the ensemble at all and just analyse it as is.

### DIFF
--- a/prody/ensemble/functions.py
+++ b/prody/ensemble/functions.py
@@ -416,7 +416,7 @@ def buildPDBEnsemble(atomics, ref=None, title='Unknown', labels=None, unmapped=N
 
     :arg superpose: if set to ``'iter'``, :func:`.PDBEnsemble.iterpose` will be used to 
         superpose the structures, otherwise conformations will be superposed with respect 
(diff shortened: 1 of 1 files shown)


prody/ensemble/functions.py
```diff
@@ -416,7 +416,7 @@ def buildPDBEnsemble(atomics, ref=None, title='Unknown', labels=None, unmapped=N
 
     :arg superpose: if set to ``'iter'``, :func:`.PDBEnsemble.iterpose` will be used to 
         superpose the structures, otherwise conformations will be superposed with respect 
-        to the reference specified by *ref*. Default is ``'iter'``
+        to the reference specified by *ref* unless set to ``False``. Default is ``'iter'``
     :type superpose: str
```
SHZ66, line 420: add `bool` to type descriptor

jamesmkrieger: like this?

SHZ66: Yes

PR #1061, opened 2020-06-09
error fix for buildPDBEnsemble with multi-state PDB

jamesmkrieger:
Two problems are fixed here:

1. The current form accepts single structures and throws an error when it iterates through the atoms in them. Therefore, a list-like has to be checked for instead of just using `len`, which corresponds to `numAtoms` for atomgroups. Only then can we use `len` to check the length of the list-like.

2. The list is allowed to have one element if `degeneracy=False` because then you have multiple structures from one atomic so that gets included in the check too.


SHZ66:
Nice fix. I am merging this as is.

PR #1063, opened 2020-06-09
added file ext adding to loadEnsemble

jamesmkrieger:
Seeing as they're saved like that, it makes it easier to be able to load like that too.


SHZ66:
Unlike saving files where extensions can be added automatically,

jamesmkrieger:
Why is this reopened?

SHZ66:
> Why is this reopened?

I wasn't sure about the change and wanted to comment but decided to do that later, and then accidentally clicked on the close button.

jamesmkrieger:
ok. It's exactly copied from saveEnsemble so should we change it there too?

PR #1066, opened 2020-06-09
Few changes to SignDy functions

SHZ66:
- Added `trimModel` that simply slice the connectivity matrix (instead of modes as does by `sliceModel`).
- Added `calcSignaturePerturbResponse`.
- `calcENM` now accepts `numpy.ndarray` as input.
- Added `getIndices` and `setIndices` to `Ensemble` (and `PDBEnsemble`).

### DIFF
--- a/prody/dynamics/editing.py
+++ b/prody/dynamics/editing.py
@@ -19,7 +19,7 @@ if not _PY3K:
 
 __all__ = ['extendModel', 'extendMode', 'extendVector',
            'sliceMode', 'sliceModel', 'sliceModelByMask', 'sliceVector',
-           'reduceModel', 'reduceModelByMask']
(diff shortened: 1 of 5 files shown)


jamesmkrieger:
Looks good to me except why are there commented out LOGGER steps? Probably they need changing the verbosity mode instead?

SHZ66:
> Looks good to me except why are there commented out LOGGER steps? Probably they need changing the verbosity mode instead?

Because the current PRS calculation, as compared to the original random perturbation method, is quite efficient (it is almost as efficient as calculating cross-correlations and we don't measure time for that), and it results in a lot of messages saying PRS is calculated in 0 seconds for most systems. So I decided to take them out.

PR #1070, opened 2020-06-10
bug fix on color handling for showProjection

jamesmkrieger:
In the previous version, a 1D array or tuple with the same length as the number of data points would become a matrix with that length in both directions, and indict would get messed up. This was to account for arrays or tuples with RGB or RGBA values.


SHZ66 (changes requested):
Import checker from `~matplotlib.pyplot.scatter` or make it clearer in the documentation about the input format

prody/dynamics/plotting.py
```diff
@@ -2381,3 +2355,43 @@ def showTree_networkx(tree, node_size=20, node_color='red', node_shape='o',
...
+
+def checkColors(colors, num, labels, allowNumbers=False):
+    """Check colors and process them if needed"""
+
+    from matplotlib.colors import is_color_like
+
+    colors_dict = {}
+
+    if isinstance(colors, np.ndarray):
+        colors = tuple(colors)
```
SHZ66, line 2368: Why does it need a numpy array check here? I believe `is_color_like` can handle numpy arrays.

prody/dynamics/plotting.py
```diff
@@ -2381,3 +2355,43 @@ def showTree_networkx(tree, node_size=20, node_color='red', node_shape='o',
...
+
+    if is_color_like(colors) or colors is None:
+        colors = [colors] * num
+    elif isListLike(colors):
+        colors = list(colors)
+
+    if isinstance(colors, dict):
```
SHZ66, line 2375: Should this be an `elif` (for better readability)?

prody/dynamics/plotting.py
```diff
@@ -2381,3 +2355,43 @@ def showTree_networkx(tree, node_size=20, node_color='red', node_shape='o',
...
+        if labels is None:
+            raise TypeError('color must be a string or a list unless labels are provided')
+        colors_dict = colors
+        colors = [colors_dict[label] for label in labels]
+
+    if isinstance(colors, list):
+        if len(colors) != num and not is_color_like(colors):
```
SHZ66, line 2382: I don't think we need `and not is_color_like(colors)` here since if it was color-like, line 2371 would have converted it to a list of colors (of `num`) already (correct me if I am wrong).

prody/dynamics/plotting.py
```diff
@@ -2381,3 +2355,43 @@ def showTree_networkx(tree, node_size=20, node_color='red', node_shape='o',
...
+            raise ValueError('colors should have the length of the set to be colored or satisfy matplotlib color rules')
+
+        if np.any([not is_color_like(color) for color in colors]):
```
SHZ66, line 2385: Please consider using a for-loop instead of `np.any` and list comprehension for better readability. The logic seems to be a bit convoluted for what it does. Especially since there's another list comprehension in line 2388.

prody/dynamics/plotting.py
```diff
@@ -2381,3 +2355,43 @@ def showTree_networkx(tree, node_size=20, node_color='red', node_shape='o',
...
+            if not allowNumbers:
+                raise ValueError('each element of colors should satisfy matplotlib color rules')
+            elif np.any([not isinstance(color, Number) for color in colors]):
```
SHZ66, line 2388: Same as above.

prody/dynamics/plotting.py
```diff
@@ -2381,3 +2355,43 @@ def showTree_networkx(tree, node_size=20, node_color='red', node_shape='o',
...
+                raise ValueError('each element of colors should be a number or satisfy matplotlib color rules')
+
+        if len(colors) > 1 and np.any([not isinstance(color, type(colors[0])) for color in colors]):
+            raise TypeError('each element of colors should have the same type')
+
+    elif not ((allowNumbers and isinstance(colors, Number)) or is_color_like(colors)):
```
SHZ66, line 2394: Do we need another `is_color_like(colors)` check here?

SHZ66 (changes requested):

prody/dynamics/plotting.py
jamesmkrieger: This is something we've had for a long time. Yes, you're right - it can.

prody/dynamics/plotting.py
jamesmkrieger: Good point. I was thinking that this is a separate thing, but it can't be true at the same time as the others, so yes.

prody/dynamics/plotting.py
jamesmkrieger: wouldn't a for loop be much slower?

What's not clear about it?

prody/dynamics/plotting.py
jamesmkrieger: same as above too

prody/dynamics/plotting.py
jamesmkrieger: Ok, I think we just need to add the Number into the check to put it in a list of length num and then it should be checked in the list check

jamesmkrieger: Probably we still want a check for what if it's not a list from all this processing?

prody/dynamics/plotting.py
SHZ66: List comprehension is slightly faster than a for-loop for **building a list**. List comprehension isn't that much different than a for-loop in terms of performance. Here, since we don't need to build a list, the code below is faster for two reasons:
1. no need to build a list for `np.any`, which requires unnecessary memory and time for the numpy array.
2. you get an early exit when one of the `colors` is wrong, so that you don't need to check the rest of them.
3. avoided a for-loop since in line 2381.

```
for color in colors:
    if not is_color_like(color):
        if not allowNumbers:
            raise ValueError('each element of colors should satisfy matplotlib color rules')
        elif not isinstance(color, Number):
            raise ValueError('each element of colors should be a number or satisfy matplotlib color rules')
```

Regarding readability, most of time list comprehension can increase readability, but sometimes it can overcomplicate things, especially if it is used in an `if` statement.

jamesmkrieger: ok, that makes sense. Done

prody/dynamics/plotting.py
```diff
@@ -2381,3 +2355,39 @@ def showTree_networkx(tree, node_size=20, node_color='red', node_shape='o',
...
+
+        for color in colors:
+            if not is_color_like(color):
+                if not allowNumbers:
+                    raise ValueError('each element of colors should satisfy matplotlib color rules')
+                elif not isinstance(color, Number):
+                    raise ValueError('each element of colors should be a number or satisfy matplotlib color rules')
+
+            if not isinstance(color, type(colors[0])):
+                raise TypeError('each element of colors should have the same type')
```
SHZ66, line 2389: Is this check necessary? Couldn't the matplotlib function handle colors defined in different ways in the same list?

prody/dynamics/plotting.py
```diff
@@ -2381,3 +2355,39 @@ def showTree_networkx(tree, node_size=20, node_color='red', node_shape='o',
...
+
+
+def checkColors(colors, num, labels, allowNumbers=False):
+    """Check colors and process them if needed"""
+
+    from matplotlib.colors import is_color_like
+
+    colors_dict = {}
+
+    if is_color_like(colors) or colors is None or isinstance(colors, Number):
```
SHZ66, line 2367: Should `colors is None` be here? Wouldn't a list of None's fail the check down below?

prody/dynamics/plotting.py
```diff
@@ -2381,3 +2355,39 @@ def showTree_networkx(tree, node_size=20, node_color='red', node_shape='o',
...
+            if not is_color_like(color):
+                if not allowNumbers:
+                    raise ValueError('each element of colors should satisfy matplotlib color rules')
+                elif not isinstance(color, Number):
+                    raise ValueError('each element of colors should be a number or satisfy matplotlib color rules')
+
+            if not isinstance(color, type(colors[0])):
+                raise TypeError('each element of colors should have the same type')
+    else:
+        raise TypeError('colors should be a colour spec or convertible to a list of color specs')
```
SHZ66, line 2391: Nitpick: `colour spec` -> `color spec` for style consistency.

prody/dynamics/plotting.py
```diff
@@ -575,21 +563,7 @@ def showCrossProjection(ensemble, mode_x, mode_y, scale=None, *args, **kwargs):
...
-    elif isinstance(colors, dict):
-        if labels is None:
-            raise TypeError('color must be a string or a list unless labels are provided')
-        colors_dict = colors
-        colors = [colors_dict[label] for label in labels]
-    else:
-        raise TypeError('color must be a string or a list or a dict if labels are provided')
+    colors, colors_dict = checkColors(colors, num, labels)
 
     if labels is not None and len(colors_dict) == 0:
```
SHZ66, line 568: I think this section can go away and `checkColors` doesn't need to return `color_dict`.

prody/dynamics/plotting.py
```diff
@@ -2381,3 +2355,39 @@ def showTree_networkx(tree, node_size=20, node_color='red', node_shape='o',
...
+        colors_dict = colors
+        colors = [colors_dict[label] for label in labels]
+
+    if isinstance(colors, list):
+        if len(colors) != num:
+            raise ValueError('colors should have the length of the set to be colored or satisfy matplotlib color rules')
+
+        for color in colors:
+            if not is_color_like(color):
+                if not allowNumbers:
```
SHZ66, line 2383: I think if `allowNumbers`, you may need to convert the number to the color in cycle like this:
```
cycle_colors = plt.rcParams['axes.prop_cycle'].by_key()['color']
color = cycle_colors[color % len(cycle_colors)]
```

SHZ66: So the entire loop may look like this:
```
if isinstance(colors, list):
        if len(colors) != num:
            raise ValueError('colors should have the length of the set to be colored or satisfy matplotlib color rules')

        for i, color in enumerate(colors):
            if not is_color_like(color):
                if allowNumbers:
                    cycle_colors = plt.rcParams['axes.prop_cycle'].by_key()['color']
                    colors[i] = cycle_colors[color % len(cycle_colors)]
                else:
                    raise ValueError('each element of colors should satisfy matplotlib color rules')
       ....
```

prody/dynamics/plotting.py
jamesmkrieger: oh, I guess it probably could

prody/dynamics/plotting.py
jamesmkrieger: yes, it would fail so I added it down there too, because we do want to allow it

prody/dynamics/plotting.py
jamesmkrieger: done

prody/dynamics/plotting.py
jamesmkrieger: I think there's a conversion of the numbers with the color cycle somewhere else, but yes, we could maybe move it here.

prody/dynamics/plotting.py
jamesmkrieger: There is a case where we need colors_dict on line 317 where we make a line graph. We'll have to find a way to adjust that.

prody/dynamics/plotting.py
jamesmkrieger: Actually, we use matplotlib.colors.Normalize instead to link up with the cmap. This is on lines 356 and 428

This could somehow be incorporated too though

jamesmkrieger:
I've no idea whether this is even still a problem that needs fixing, but it should be checked anyway

PR #1086, opened 2020-06-17
added option for showOverlap to show diagonals from NMA or ModeSet co…

jamesmkrieger:
…mparisons

```
plt.figure(); showOverlap(anm_mod2, anm, diag=True);
```
yields adding_100_to_gamma_for_15-30A_apart_cysteines_3o21_CD_ANM_abs_diag_overlaps, which is the diagonal of 
adding_100_to_gamma_for_15-30A_apart_cysteines_3o21_CD_ANM_abs_overlaps

### DIFF
--- a/prody/dynamics/plotting.py
+++ b/prody/dynamics/plotting.py
@@ -820,8 +820,8 @@ def showContactMap(enm, **kwargs):
 def showOverlap(mode, modes, *args, **kwargs):
     """Show overlap :func:`~matplotlib.pyplot.bar`.
 
-    :arg mode: a single mode/vector
-    :type mode: :class:`.Mode`, :class:`.Vector`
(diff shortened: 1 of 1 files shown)


prody/dynamics/plotting.py
```diff
@@ -833,13 +833,22 @@ def showOverlap(mode, modes, *args, **kwargs):
     if SETTINGS['auto_show']:
         plt.figure()
 
-    if not isinstance(mode, (Mode, Vector)):
-        raise TypeError('mode must be Mode or Vector, not {0}'
-                        .format(type(mode)))
+    diag = kwargs.pop('diag', False)
```
SHZ66, line 836: `diag` option might not be necessary.

PR #1087, opened 2020-06-18
Tidying up emdmap code

jamesmkrieger:
The documentation was wrong, the make_nodes parameter was unnecessary, and the `_parseEMDLines`/`parseEMDStream` pair was a debugging nightmare. Now this should be better.


jamesmkrieger:
I made it go back to returning `atomgroup, emd` like we have `atomgroup, header` in `parsePDB`. I don't know why we keep having to do that but now there's only one place that controls it in the combined parseEMDStream so we shouldn't keep having to worry about it.

PR #1085, opened 2020-06-16
added function printAtomicMatrix

jamesmkrieger:
Example:
```
Python 3.7.1 (default, Dec 14 2018, 13:28:58) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.13.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from prody import * 
   ...: import numpy as np 
   ...: import sys 
   ...:  
   ...: ag = parsePDB('3hsyA', subset='ca') 
   ...: cys = ag.select('resname CYS and ca') 
   ...:  
   ...: gnm = GNM('3hsy') 
   ...: gnm.buildKirchhoff(ag) 
   ...:  
   ...: K = gnm.getKirchhoff() 
   ...: cys_k = K[cys._indices, :][:, cys._indices]                                                               

In [2]: printAtomicMatrix(cys_k, cys)                                                                             
        CYS57   CYS83   CYS190  CYS309
CYS57   17.0     0.0     0.0    -1.0
CYS83    0.0    21.0     0.0     0.0
CYS190   0.0     0.0    21.0     0.0
CYS309  -1.0     0.0     0.0    18.0
```

PR #1098, opened 2020-07-14
Added calcModeClusters method to utilities/catchall.py

bwingert:


jamesmkrieger:
You also need to add this to all (breaking line 44) to make it useable.

prody/dynamics/signature.py
```diff
@@ -1001,6 +1007,106 @@ def _getEnsembleENMs(ensemble, **kwargs):
...
+                    the values themselves.
+    :type method: str
+
+    :arg cutoff: only used if *method* is set to **spectral**. is the value used for 
+                    generating the Kirchoff matrix to use for generating clusters.
+    :type cutoff: float
+    """
+
+    # Check inputs to make sure are of valid types/values
+    if not isinstance(similarities, (np.array, np.ndarray, pd.DataFrame)):
```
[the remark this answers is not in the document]
jamesmkrieger: Your code gave the following error before I removed np.array:

```
In [2]: rmsd_cluster_labels = calcModeClusters(rmsd_matrix)                                                                           
TypeError                                 Traceback (most recent call last)
<ipython-input-2-2a481e078ff0> in <module>
----> 1 rmsd_cluster_labels = calcModeClusters(rmsd_matrix)

~/Documents/code/ProDy/prody/dynamics/signature.py in calcModeClusters(similarities, clusters, linkage, method, cutoff, **kwargs)
   1037 
   1038     # Check inputs to make sure are of valid types/values
-> 1039     if not isinstance(similarities, (np.array, np.ndarray, pd.DataFrame)):
   1040         raise TypeError('similarities should be an instance of ndarray or DataFrame')
   1041 

TypeError: isinstance() arg 2 must be a type or tuple of types
```

bwingert:
Moved calcModeClusters to utilities/catchall.py. Made requested changes so doesn't require Pandas as a dependency, though will accept a DataFrame if given as an input.

prody/utilities/catchall.py
```diff
@@ -7,13 +7,125 @@
...
+        from sklearn.metrics import silhouette_score
+        from sklearn.manifold import TSNE
+    except ImportError:
+        raise ImportError('need sklearn module')
+
+    # Check inputs to make sure are of valid types/values
+    if not isinstance(similarities, np.ndarray):
+        try:
+            import pandas as pd
+            if not isinstance(similarities, np.ndarray):
```
jamesmkrieger, line 60: This should be pd.DataFrame

prody/utilities/catchall.py
```diff
@@ -7,13 +7,125 @@
...
+            raise ValueError('clusters must be a positive integer')
+        elif n_clusters > similarities.shape[0]:
+            raise ValueError('clusters can\'t be longer than similarities matrix')
+        nclusts = range(n_clusters,n_clusters+1)
+    else:
+        nclusts = range(2,10,1)
+
+    if linkage != 'all':
+        if not isinstance(method, str):
+            raise TypeError('linkage must be an instance of str')
```
jamesmkrieger, line 82: It might be good to let the user provide a list-like of methods

prody/utilities/catchall.py
bwingert: Fixed

prody/utilities/catchall.py
bwingert: That's a good idea, I've added that as an option.

prody/utilities/catchall.py
```diff
@@ -78,10 +87,18 @@ def calcModeClusters(similarities, n_clusters=0, linkage='all', method='tsne', c
         nclusts = range(2,10,1)
 
     if linkage != 'all':
-        if not isinstance(method, str):
-            raise TypeError('linkage must be an instance of str')
-        if method not in ['ward', 'average', 'complete', 'single']:
-            raise ValueError('linkage must be either \'ward\', \'average\', \'complete\', or \'single\'')
+        if isinstance(linkage, list):
+            if len(linkage) > 4:
```
jamesmkrieger, line 91: I'd probably do this after the values checking. That way you're checking if something is repeated. It's not relevant how long the list is if there's a wrong value in it.

prody/utilities/catchall.py
```diff
@@ -78,10 +87,18 @@ def calcModeClusters(similarities, n_clusters=0, linkage='all', method='tsne', c
...
+                raise ValueError('linkage must be one or more of: \'ward\', \'average\', \'complete\', or \'single\'')
+            for val in method:
```
jamesmkrieger, line 93: shouldn't this be linkage?

prody/utilities/catchall.py
```diff
@@ -78,10 +87,18 @@ def calcModeClusters(similarities, n_clusters=0, linkage='all', method='tsne', c
...
+                if val.lower() not in ['ward', 'average', 'complete', 'single']:
+                    raise ValueError('linkage must be one or more of: \'ward\', \'average\', \'complete\', or \'single\'')
+                linkages = [ x.lower() for x in linkage ]
+        elif not isinstance(linkage, str):
+            raise TypeError('linkage must be an instance of str or list of strs')
+
+        if linkage not in ['ward', 'average', 'complete', 'single']:
```
jamesmkrieger, line 100: This will be tripped if you have a list there.

prody/utilities/catchall.py
```diff
@@ -29,9 +29,9 @@ def calcModeClusters(similarities, n_clusters=0, linkage='all', method='tsne', c
     :type n_clusters: int
 
     :arg linkage: if **all**, will test all linkage types (ward, average, complete,
-                    single). Otherwise will use only the one given as input. Default is
+                    single). Otherwise will use only the one(s) given as input. Default is
                     **all**.
-    :type linkage: str
+    :type linkage: str, list
```
jamesmkrieger, line 34: I'd suggest adding numpy ndarray and tuple here if you use isListLike like I suggested.

prody/utilities/catchall.py
bwingert: Yes, fixed

prody/utilities/catchall.py
```diff
@@ -63,6 +63,8 @@ def calcModeClusters(similarities, n_clusters=0, linkage='all', method='tsne', c
         
 
     # Check inputs to make sure are of valid types/values
+    from prody.utilities import isListLike
```
jamesmkrieger, line 66: This should be added to line 6 not here. You don't want to import `prody.utilities` inside `prody.utilities.catchall` too. `.misctools` means `misctools` from the same module (`prody.utilities.misctools`).

prody/utilities/catchall.py
jamesmkrieger: This isn't resolved yet. You need to indent this line and the next one to make it be part of the str block.

prody/utilities/catchall.py
jamesmkrieger: You can see #1085 where I've done that

prody/utilities/catchall.py
```diff
@@ -63,14 +63,12 @@ def calcModeClusters(similarities, n_clusters=0, linkage='all', method='tsne', c
...
     # Check inputs to make sure are of valid types/values
-    from prody.utilities import isListLike
-
     if not isinstance(similarities, np.ndarray):
         try:
             import pandas as pd
             if not isinstance(similarities, pd.DataFrame):
                 raise TypeError('similarities should be an instance of ndarray or DataFrame')
-        except:
+        except ImportError:
```
jamesmkrieger, line 71: That doesn't make sense. If there's an ImportError then it's not a TypeError. The TypeError goes with the np.ndarray check and should be back-indented. Something else should go here such as `pass`.

prody/utilities/catchall.py
jamesmkrieger: Ok this is good now

prody/utilities/catchall.py
jamesmkrieger: It still might be easier to just remove pandas and have the user convert to numpy like they do for everything else. You could also have a variable marking if pandas is imported and do the type checking with that rather than coupled to the importing maybe.

prody/utilities/catchall.py
```diff
@@ -3,17 +3,151 @@
...
-from .misctools import addEnds, interpY, index
+from .misctools import addEnds, interpY, index, isListLike
 from .checkers import checkCoords
 from Bio.Phylo.BaseTree import Tree, Clade
 
+
 __all__ = ['calcTree', 'clusterMatrix', 'showLines', 'showMatrix', 
            'reorderMatrix', 'findSubgroups', 'getCoords',  
-           'getLinkage', 'getTreeFromLinkage']
+           'getLinkage', 'getTreeFromLinkage', 'calcModeClusters']
```
SHZ66, line 13: Change the function's name here also accordingly, e.g. `clusterSubfamilies`.

prody/utilities/catchall.py
SHZ66: > It still might be easier to just remove pandas and have the user convert to numpy like they do for everything else. You could also have a variable marking if pandas is imported and do the type checking with that rather than coupled to the importing maybe.

Yes, I second this.

jamesmkrieger (approved):

jamesmkrieger:
I've fixed these problems using the web interface now so I'm going to merge this.

PR #1107, opened 2020-08-04
contacts docs fix

jamesmkrieger:
There was an incomplete sentence before that didn't make sense.

### DIFF
--- a/prody/measure/contacts.py
+++ b/prody/measure/contacts.py
@@ -20,7 +20,7 @@ class Contacts(object):
         
         :arg unitcell: orthorhombic unitcell dimension array with shape 
                         ``(3,)`` for KDTree. Default is **None**.
-        :type unitcell: :class:`numpy.ndarray`"""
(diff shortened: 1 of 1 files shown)


prody/measure/contacts.py
```diff
@@ -16,8 +16,11 @@ class Contacts(object):
     of instantiation."""
 
     def __init__(self, atoms, unitcell=None):
-        """*atoms* must be an :class:`.Atomic` instance.  When an orthorhombic
-        *unitcell* array is given"""
+        """*atoms* must be an :class:`.Atomic` instance.  
+        
+        :arg unitcell: orthorhombic unitcell dimension array with shape 
+                        ``(3,)`` for KDTree. Default is **None**.
+        :type unitcell: :class:`numpy.ndarray`"""
```
SHZ66, line 23: ```:type unitcell: :class:`~numpy.ndarray`"""```

jamesmkrieger: Oh yeah. Thanks

PR #1117, opened 2020-08-13
Bug fixes

jamesmkrieger:
Most of these came from commits found in other forks of ProDy including https://github.com/Wyss/ProDy/, https://github.com/Python3pkg/ProDy/ and https://github.com/T-B-F/ProDy/


jamesmkrieger:
These should probably be checked better before merging.

jamesmkrieger:
I've tested the prody apps and they all work so we should be good. The evol apps still have some problems but this shouldn't affect them. Therefore I'm happy to merge this.

Issue #1119, opened 2020-08-13
Selection bug for negative residue numbers


jamesmkrieger:
I have a structure that starts with residue -15 but we can't select with negative residue numbers:

```
(prody) Ians-iMac:ProDy jkrieger$ ipython
Python 3.7.8 | packaged by conda-forge | (default, Jul 31 2020, 02:37:09) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.17.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from prody import *

In [2]: ag = parsePDB('2vsv')
@> PDB file is found in the local folder (/Users/jkrieger/.../2vsv.pdb.gz).
@> 1631 atoms and 1 coordinate set(s) were parsed in 0.04s.

In [3]: ag.select('resnum -15')
---------------------------------------------------------------------------
UFuncTypeError                            Traceback (most recent call last)
<ipython-input-3-48af887a9e84> in <module>
----> 1 ag.select('resnum -15')

~/Documents/codes/ProDy/prody/atomic/atomic.py in select(self, selstr, **kwargs)
    232         documentation for details and usage examples."""
    233 
--> 234         return SELECT.select(self, selstr, **kwargs)
    235 
    236     def getTitle(self):

~/Documents/codes/ProDy/prody/atomic/select.py in select(self, atoms, selstr, **kwargs)
    884 
    885         self._selstr = selstr
--> 886         indices = self.getIndices(atoms, selstr, **kwargs)
    887 
    888         self._kwargs = None

~/Documents/codes/ProDy/prody/atomic/select.py in getIndices(self, atoms, selstr, **kwargs)
    942                                      'string', [ss])
    943         else:
--> 944             torf = self.getBoolArray(atoms, selstr, **kwargs)
    945             return torf.nonzero()[0]
    946 

~/Documents/codes/ProDy/prody/atomic/select.py in getBoolArray(self, atoms, selstr, **kwargs)
    994         try:
    995             parser = self._getParser(selstr)
--> 996             tokens = parser(selstr, parseAll=True)
    997         except pp.ParseException as err:
    998             self._parsers.pop(self._parser, None)

/opt/anaconda3/envs/prody/lib/python3.7/site-packages/pyparsing.py in parseString(self, instring, parseAll)
   1941             instring = instring.expandtabs()
   1942         try:
-> 1943             loc, tokens = self._parse(instring, 0)
   1944             if parseAll:
[...]

PR #1120, opened 2020-08-13
current bug fix for negative resnum selections

jamesmkrieger:
This works as a fix for #1119 but at the expense of being able to do subtraction as part of selection strings. I don't think this is a big loss but others may disagree.


jamesmkrieger:
The tests are failing because there is a test for doing just that so I guess I'll have to go back to the drawing board.

PR #1143, opened 2020-09-08
cleaned up unneeded/unused definitions

jamesmkrieger:
SELECT and isSelectionMacro are defined in atomic __init__.py and prody __init__.py

ParseException wasn't necessary to import separately.


jamesmkrieger:
I guess this should maybe be tested a little more

PR #1152, opened 2020-09-14
Simplified .travis.yml

jamesmkrieger:


simonbray:
It would be good to run the conda tests for multiple python versions as well (3.6 to 3.8)

jamesmkrieger:
Good point. It would be good to run conda and non-conda tests for multiple versions. Do you know how to do that?

jamesmkrieger:
I guess I'll look into Travis CI docs more

simonbray:
> Do you know how to do that?

Not without going into the docs sorry

jamesmkrieger:
> > Do you know how to do that?
> 
> Not without going into the docs sorry

Ok. Thanks

simonbray:
Actually, is this necessary? The conda-forge CI tests all the builds it creates anyway. Not sure it makes sense for these tests to be run every time a PR is opened.

jamesmkrieger:
> Actually, is this necessary? The conda-forge CI tests all the builds it creates anyway. Not sure it makes sense for these tests to be run every time a PR is opened.

Good point. I'll use this PR to simplify the .travis.yml instead then.

PR #1166, opened 2020-09-25
fixed showOverlap and showCumulOverlap for one mode

jamesmkrieger:
This way we can use diag=True for the diagonal of an overlap matrix between two sets of modes with einsteinsum or diag=False for overlaps between one mode and a set of modes.


jamesmkrieger:
This now works for showOverlap of an ANM with a defvec:



I could also use it for the diagonal of the overlap matrix between two ANM objects:

PR #1176, opened 2020-09-28
A few fixes to alignSequenceToMSA

yaz62:


prody/sequence/analysis.py
```diff
@@ -992,9 +993,15 @@ def alignSequenceToMSA(seq, msa, **kwargs):
             ag = seq.select('chain {0}'.format(chain))
```
jamesmkrieger, line 993: We should probably have a try except here. Someone could provide a string that isn't valid for selecting chains.

prody/sequence/analysis.py
```diff
@@ -992,9 +993,15 @@ def alignSequenceToMSA(seq, msa, **kwargs):
...
-            raise TypeError('chain should be a string or **None**')
 
+            chids = ag.getChids()
+            if len(unique(chids)) > 1:
+                LOGGER.warn('%s consists of multiple chains. Please consider selecting one chain'%(seq.getTitle()))
+        else:
+            raise TypeError('chain should be a string or None')
+        
+        if ag is None:
+            raise ValueError('seq may be None or invalid chain ID')
```
jamesmkrieger, line 1004: This is unclear. Instead of the second part, how about `chain may be invalid`?

prody/sequence/analysis.py
```diff
@@ -992,9 +993,15 @@ def alignSequenceToMSA(seq, msa, **kwargs):
...
         sequence = ag.select('ca').getSequence()
 
     elif isinstance(seq, Sequence):
```
jamesmkrieger, line 1007: We also need an `elif` for seq being a string or we need to update the error. Probably accepting strings is a good idea.

jamesmkrieger (changes requested):
I just thought of a few improvements

PR #1184, opened 2020-10-02
made refineMSA update label after structure refinement

jamesmkrieger:
Otherwise the label start and end don't match the MSA anymore and pickSequence and getResnums have problems.


jamesmkrieger:
closes #772 once the last two issues are worked out

Issue #1123, opened 2020-08-17
Add interoperability with QuartataWeb, DrugBank and STITCH


jamesmkrieger:
Probably the best way to do this would be to push jobs and pull results like we do for Dali. 

These resources all provide access to different data so it's probably a good idea to have functions for all of them. An adapter may be a good way to do that too.

QuartataWeb could be a good one to start with because it provides DrugBank and STITCH compounds as well as predictions from probabilistic matrix factorisation.

jamesmkrieger:
QuartataWeb may be a little difficult because we won't be able to predict the job ID. 

ChEMBL would be a good one to include as it has https://github.com/chembl/chembl_webresource_client already, which we could link up with. 

For PubChem, there is is a REST API (see https://pubchemdocs.ncbi.nlm.nih.gov/pug-rest).

jamesmkrieger:
DrugBank also has a REST API, but they need to be contacted to get an authentication key.

jamesmkrieger:
We now don't need a job ID for QuartataWeb as we use it from a browser and have access to the URL. We can also get to those compounds in DrugBank and PubChem from there although there are probably better ways.

jamesmkrieger:
http://files.docking.org/ provides files for ZINC etc.

jamesmkrieger:
http://www.rcsb.org/pdb/software/rest.do would be good too

jamesmkrieger:
The QuartataWeb part is done and merged. It now a module of the `database` subpackage.

jamesmkrieger:
> The QuartataWeb part is done and merged. It now a module of the `database` subpackage.

It could still do with more functionality though as currently it can only download known chemicals I think.

PR #1208, opened 2020-11-18
added QuartataWebBrowser

jamesmkrieger:
This can now navigate QuartataWeb. Downloading and parsing the results comes next.

### DIFF
--- a/prody/database/__init__.py
+++ b/prody/database/__init__.py
@@ -38,3 +38,7 @@ __all__.extend(emsurfer.__all__)
 from . import goa
 from .goa import *
 __all__.extend(goa.__all__)
+
+from . import quartataweb
+from .quartataweb import *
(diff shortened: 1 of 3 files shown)


jamesmkrieger:
This is a start towards #1123

jamesmkrieger:
It can now parse chemicals for query protein. Other things still need parsing too.

jamesmkrieger:
Together with the 2D similarity function from #1115, we can do the following:

```
from prody import *
plt.ion()

rec = searchQuartataWeb(data_source='STITCH', data=['','furin'])

raw_sim_mat = calc2DSimilarityMatrix(rec.getSMILESList())

plt.figure()
showMatrix(raw_sim_mat)
plt.savefig('furin_raw_2D_sim_mat')
```
furin_raw_2D_sim_mat

PR #1210, opened 2020-11-20
Improvements to quartata and minor fixes

jamesmkrieger:
- Quartata classes and functions now have a filename option like blastPDB
- filter function in quartata and dali records keeps the order by not using set


jamesmkrieger:
These changes have been moved to the compounds pull request (#1115).

PR #1222, opened 2020-12-21
moved IndexFormatter into ProDy before matplotlib removes it

jamesmkrieger:
The IndexFormatter is needed by too many functions to lose it and matplotlib has had deprecation warnings for it for a while.


jamesmkrieger:
We may want to find another way to do this so that matplotlib doesn't become a dependency, but we provide so much plotting functionality that it's probably not an issue.

PR #1224, opened 2020-12-21
moved IndexFormatter into ProDy before matplotlib removes it

jamesmkrieger:
The IndexFormatter is needed by too many functions to lose it and matplotlib has had deprecation warnings for it for a while.

### DIFF
--- a/prody/database/charmmgui.py
+++ b/prody/database/charmmgui.py
@@ -24,45 +24,43 @@ __all__ = ['CharmmGUIBrowser']
 
 
 class CharmmGUIBrowser(object):
-    """Class for running CHARMM-GUI in a browser"""
+    """Class for running CHARMM-GUI in a browser
 
+    :arg fname: filename for starting PDB file. A PDB ID can be provided and ProDy will fetch the PDB file first.
+    :type fname: str
+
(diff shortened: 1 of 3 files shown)


jamesmkrieger:
it works:
```
Python 3.7.8 | packaged by conda-forge | (default, Jul 31 2020, 01:53:57) [MSC v.1916 64 bit (AMD64)]
Type 'copyright', 'credits' or 'license' for more information
IPython 7.17.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from prody import *

In [2]: import numpy as np

In [3]: import matplotlib.pyplot as plt

In [4]: plt.ion()

In [5]: m = np.random.randint(0, 10, size=(200,200))

In [6]: showMatrix(m)
Out[6]: 
(<matplotlib.image.AxesImage at 0x1d0d6a22cc8>,
 [],
 <matplotlib.colorbar.Colorbar at 0x1d0d4084288>)

In [7]: plt.savefig('showMatrix_random_test')

In [8]: plt.close()
```
C:\Users\james\code\ProDy\showMatrix_random_test.png

Still need to fix doc string.

PR #1238, opened 2020-12-30
Minor fixes in starfile

SHZ66:
Comparing string literals with identity operator "is" is not a good practice and may cause unexpected bugs. The string class is in a sweet spot between value-type and object-type variables, and in Python, it is treated in a mixed fashion. Using "!=" here is the intended comparison.


jamesmkrieger:
Ok. Thanks

PR #1258, opened 2021-01-05
Restructured aANM

SHZ66:
Docs changes will come in the next pull request.

### DIFF
--- a/prody/dynamics/adaptive.py
+++ b/prody/dynamics/adaptive.py
@@ -1,140 +1,96 @@
 # -*- coding: utf-8 -*-
 """This module defines functions for performing adaptive ANM."""
 
-from prody.atomic.atomic import Atomic
+from prody.atomic import Atomic, AtomMap
 import time
 from numbers import Integral, Number
 import numpy as np
 
 from prody import LOGGER
-from prody.utilities import getCoords
+from prody.utilities import getCoords, importLA
 from prody.measure import calcRMSD, calcDistance, superpose
 from prody.ensemble import Ensemble
 
 from .functions import calcENM
 from .modeset import ModeSet
 
+__all__ = ['calcAdaptiveANM', 'AANM_ONEWAY', 'AANM_ALTERNATING', 'AANM_BOTHWAYS', 'AANM_DEFAULT']
 
-__all__ = ['AdaptiveANM', 'runAdaptiveANM', 'runOneWayAdaptiveANM', 'runBothWaysAdaptiveANM']
+AANM_ONEWAY = 0
+AANM_ALTERNATING = 1
+AANM_BOTHWAYS = 2
 
+AANM_DEFAULT = AANM_ONEWAY
 
-class AdaptiveANM(object):
-    """Class for initialising adaptive ANM analysis of proteins ([ZY09]_).
+norm = importLA().norm
 
-    This implementation differs from the original one in that it sorts the 
-    modes by overlap prior to cumulative overlap calculations for efficiency.
+def checkInput(structureA, structureB, **kwargs):
+    weights = kwargs.pop('weights', 1.)
 
-    .. [ZY09] Zheng Yang, Peter Májek, Ivet Bahar. Allosteric Transitions of 
-              Supramolecular Systems Explored by Network Models: Application to 
-              Chaperonin GroEL. *PLOS Comp Biol* **2009** 40:512-524.
+    coordsA = getCoords(structureA)
+    if isinstance(structureA, Atomic):
+        title = structureA.getTitle()
+        atoms = structureA
+    else:
+        title = None
+        atoms = None
 
-    :arg structureA: starting structure for the transition
-    :type structureA: :class:`.Atomic`, :class:`~numpy.ndarray`
-
-    :arg structureB: starting structure for the transition
-    :type structureB: :class:`.Atomic`, :class:`~numpy.ndarray`
-    """
-
-    def __init__(self, structureA, structureB, **kwargs):
-
-        self.coordsA = getCoords(structureA)
-        self.coordsA_0 = self.coordsA
-        if isinstance(structureA, Atomic):
-            self.titleA = structureA.getTitle()
-        else:
-            self.titleA = 'structure A'
-
-        self.coordsB = getCoords(structureB)
-        self.coordsB, _ = superpose(self.coordsB, self.coordsA)
-        self.coordsB_0 = self.coordsB
+    coordsB = getCoords(structureB)
+    
+    if title is None:
         if isinstance(structureB, Atomic):
-            self.titleB = structureB.getTitle()
+            title = structureB.getTitle()
+            atoms = structureB
         else:
-            self.titleB = 'structure B'
+            title = 'Unknown'
+            atoms = None
 
-        if self.coordsA.shape != self.coordsB.shape:
-            raise ValueError('structures should have the same number of atoms')
+    weightsA = structureA.getFlags("mapped") if isinstance(structureA, AtomMap) else 1.
+    weightsB = structureB.getFlags("mapped") if isinstance(structureB, AtomMap) else 1.
+    weights = weights * weightsA * weightsB
 
-        self.n_atoms = self.coordsA.shape[0]
+    if np.isscalar(weights):
+        weights = None
 
-        rmsd = calcRMSD(self.coordsA, self.coordsB)
-        self.rmsds = [rmsd]
-        self.dList = []
-        self.numSteps = 0
+    coordsA, _ = superpose(coordsA, coordsB, weights)
+    rmsd = calcRMSD(coordsA, coordsB, weights)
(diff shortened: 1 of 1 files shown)


prody/dynamics/adaptive.py
```diff
@@ -1,140 +1,96 @@
...
-        self.ensemble = Ensemble('combined trajectory')
-        self.ensemble.setCoords(self.coordsA)
-
-        self.rmsd_diff_cutoff = kwargs.get('rmsd_diff_cutoff', 0.05)
-        self.target_rmsd = kwargs.get('target_rmsd', 1.0)
-
-        LOGGER.info(
-            'Initialised Adaptive ANM with RMSD {:4.3f}\n'.format(rmsd))
+def getTitle(structure, def_title='structure'):
+    if isinstance(structure, Atomic):
```
jamesmkrieger, line 62: could use has_attr but this is fine too

prody/dynamics/adaptive.py
```diff
@@ -1,140 +1,96 @@
...
+        title = def_title
 
+    return title
 
-def _runStep(structureA, structureB, **kwargs):
+def calcStep(initial, target, n_modes, ensemble, defvecs, rmsds, callback_func=None, **kwargs):
     """Run a single step of adaptive ANM. 
-    Modes will be calculated for *structureA* and the subset with 
+    Modes will be calculated for *initial* and the subset with 
     a cumulative overlap above a threshold defined by *Fmin*
```
jamesmkrieger, line 72: We should maybe say squared cumulative overlap

prody/dynamics/adaptive.py
```diff
@@ -1,140 +1,96 @@
...
 
     Fmin_max = kwargs.get('Fmin_max', 0.6)
     resetFmin = kwargs.get('resetFmin', False)
 
-    dList = kwargs.get('dList', [])
-    numModesList = kwargs.get('numModesList', [])
+    weights = ensemble.getWeights()
+    if weights is not None:
+        weights = weights.flatten()
+    # coords_init, _ = superpose(initial, target, weights) # we should keep this off otherwise RMSD calculations are off
```
jamesmkrieger, line 89: Yes, this can be removed because we have it in calcStep

prody/dynamics/adaptive.py
```diff
@@ -146,135 +102,135 @@ def _runStep(structureA, structureB, **kwargs):
...
-    if not isinstance(n_modes, Integral):
-        raise TypeError('n_modes should be an integer')
-
     if n_modes > maxModes:
         n_modes = maxModes
 
-    anmA, _ = calcENM(coordsA, n_modes=n_modes, **kwargs)
+    mask = None if weights is None else weights != 0
+    anm, _ = calcENM(coords_init, select=mask, mask=mask, 
+                     model='anm', trim='trim', n_modes=n_modes,
```
jamesmkrieger, line 110: Probably better to use reduce and take the non-overlapping part as environment by default. Better still would be to give the user control over this parameter.

prody/dynamics/adaptive.py
```diff
@@ -146,135 +102,135 @@ def _runStep(structureA, structureB, **kwargs):
...
-    1. difference between *rmsds* from previous step to current < *rmsd_diff_cutoff*
+    1. difference between *rmsds* from previous step to current < *min_rmsd_diff*
     2. Current rmsd < *target_rmsd*
-    3. A node in *coordsA* or *coordsB* gets disconnected from another by > *cutoff*
+    3. A node in *coords* gets disconnected from another by > *cutoff*
 
     :arg rmsds: a list of RMSDs from Adaptive ANM
     :type rmsds: list
 
     :arg coordsA: coordinate set A for checking disconnections
```
jamesmkrieger, line 200: this should be coords not coordsA and the description needs to change accordingly

prody/dynamics/adaptive.py
```diff
@@ -369,71 +325,57 @@ def runAlternatingAdaptiveANM(structureA, structureB, n_steps, **kwargs):
...
+
+        if n_modes == 0:
+            LOGGER.report('Alternating Adaptive ANM converged in %.2fs.', '_prody_calcAdaptiveANM')
             break
 
-    for coordset in self.ensembleA.getCoordsets():
-        self.ensemble.addCoordset(coordset)
-    for coordset in reversed(self.ensembleB.getCoordsets()):
-        self.ensemble.addCoordset(coordset)
+        LOGGER.info('\nStarting cycle {0} with structure {1}'.format(n+1, getTitle(structureB, 'structure B')))
```
jamesmkrieger, line 361: This should be above so you report starting cycle 1 (n=0) too. 

Might also be good to have this as debug not info.

prody/dynamics/adaptive.py
```diff
@@ -451,80 +393,59 @@ def runBothWaysAdaptiveANM(structureA, structureB, n_steps, **kwargs):
...
-            # That way the original object is back to normal
-            #self.coordsA = self.coordsA_0
-            # That way the original object is back to normal
-            #self.coordsB = self.coordsB_0
-            LOGGER.warn('The part starting from structure A converged.')
-            break
 
-    LOGGER.info('\n\nStarting from structure B ({0})'.format(self.titleB))
+        if n_modes == 0:
+            LOGGER.report('Alternating Adaptive ANM converged in %.2fs.', '_prody_calcAdaptiveANM')
```
jamesmkrieger, line 426: Add that this is from structure A and the one below is from structure B.

prody/dynamics/adaptive.py
```diff
@@ -451,80 +393,59 @@ def runBothWaysAdaptiveANM(structureA, structureB, n_steps, **kwargs):
...
+            break
+    
+    n = 0
```
jamesmkrieger, line 429: I was expecting this to keep counting up rather than resetting so the total number of steps from both is used as a convergence criterion. This may work too. I guess it needs documenting carefully.

jamesmkrieger (changes requested):
Overall looks good. Just a few minor things

prody/dynamics/adaptive.py
SHZ66: Yes, but `isinstance` may be better in this case because it is more consistent with the function above.

prody/dynamics/adaptive.py
SHZ66: I plan to refresh the docs in the next PR but this is a good note

prody/dynamics/adaptive.py
jamesmkrieger: ok

prody/dynamics/adaptive.py
jamesmkrieger: ok

prody/dynamics/adaptive.py
SHZ66: This needs to be `trim` because the non-overlapping parts are filled by dummies atoms in either structure, and the dummies could have the same coordinates, e.g. `[0, 0, 0]`. And this will cause a NaN in the Hessian. `trim` is used to get rid of these NaN. But if we use reduceModel, these NaN will cause an error.

prody/dynamics/adaptive.py
SHZ66: I simply removed the one for structureA. One message for the overall time should be good enough

prody/dynamics/adaptive.py
SHZ66: The number of steps is not used as a convergence criterion currently. This needs to be reset otherwise `n<n_steps` can be `False` and structure B never got run. `n` is an internal variable for controlling the loop logic anyway.

prody/dynamics/adaptive.py
jamesmkrieger: Ok. Sounds good.

prody/dynamics/adaptive.py
jamesmkrieger: Yep, that works too

prody/dynamics/adaptive.py
jamesmkrieger: Perhaps you could change n_steps to reflect the steps that already took place.

prody/dynamics/adaptive.py
SHZ66: After a second thought, I decided to make mask and weights separate and the dummies will still be trimmed, while their counterparts in the other structures are calculated as usual. Then the weights are used separately to calculate RMSD, superpose, and the deformation vector.

prody/dynamics/adaptive.py
SHZ66: I think resetting `n` is simpler.

jamesmkrieger: No, I meant if you change `n_steps` then it reflects the reset to n. The new `n_steps` for B would be `n_steps_in - n_steps` for A if you reset `n`. If you don't reset `n` then `n_steps` stays as `n_steps_in`.

jamesmkrieger (approved):

PR #1261, opened 2021-01-05
Major fixes to aANM

SHZ66:
Now aANM code should reproduce the results in the paper.


SHZ66:
The changes are tested by reproducing the paper's results. Further changes should be minor and made carefully.

PR #1265, opened 2021-01-11
Removed selstrs from PDBEnsemble

SHZ66:
`PDBEnsemble._selstrs` is removed for several reasons:

- It causes a bug in `saveEnsemble` when saving `Ensemble` objects. This is a `PDBEnsemble` only attribute but it was saved as if it were an attribute in the base class. 
- It was implemented incompletely. This attribute was not addressed at all in editing operators such as `__add__` and `__getitem__`. Future additions of new attributes should be made carefully as there are many aspects (e.g. loading, saving, and editing) of the class to be considered before making the decision to add an attribute.
- It is not very useful. Atomic subsets can be created without a selection string, namely AtomMap, which is the subset class used for creating the `PDBEnsemble`. Even if there is a selection string, the users cannot use it unless they have the access to the original `AtomGroup` instance, in which case they should just save the atomic subset objects (by themselves, not by `PDBEnsemble`) as well. 

With that being said, we might consider adding a callback function in `buildPDBEnsemble` in the future so that users can have access to the AtomMap or other local objects for debugging purposes. But that requires some design decisions to be made and tutorial changes, so this will be a future project (#1266).

### DIFF
--- a/prody/ensemble/functions.py
+++ b/prody/ensemble/functions.py
@@ -60,11 +60,6 @@ def saveEnsemble(ensemble, filename=None, **kwargs):
         attr_dict['_atoms'] = np.array([atoms, None], 
                                         dtype=object)
 
-    selstrs = dict_['_selstrs']
-    if selstrs is not None:
-        attr_dict['_selstrs'] = np.array([selstrs, None], 
-                                          dtype=object)
-
     data = dict_['_data']
     if len(data):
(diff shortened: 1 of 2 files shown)


jamesmkrieger (approved):
I agree that selstrs was causing problems without doing anything helpful and should be removed

PR #1267, opened 2021-01-14
added thresholdMap and max_cutoff

jamesmkrieger:
We can now threshold in both directions and there is now a function to do that after loading too. I also added some type checkers.


jamesmkrieger:
I also made the thresholding more efficient following the implementation in TEMPy that multiplies by boolean arrays.

jamesmkrieger:
The checks aren’t quite right

jamesmkrieger:
This now works. Test code was essentially the following:
```
import prody
threshold = 0.625
pd_emd2 = prody.parseEMD('1961', cutoff=threshold)
prody.writeEMD('1961-%3.2f.map' % threshold, pd_emd2)
```

PR #1268, opened 2021-01-14
writeNMD bug fix

jamesmkrieger:
There might not be numModes() so this works more of the time.


jamesmkrieger:
Actually, this doesn't make sense for ModeSet

Issue #746, opened 2018-08-19
PCA job submission error in NMWiz


JazminOzsvar:
Hi all,

I'm currently trying to work through the tutorial files for PCA ensemble analysis on the ProDy website. I'm using Python 2.7.10 and VMD 1.9.3. After loading my files as per the tutorial clicking "submit", NMWiz gives me this error:

@> Superposing  [ 97%] 1s
                         
@> Superposing  [ 98%] 1s
                         
@> Superposing  [ 99%] 1s
                         
@> Superposing  [100%] 0s

while executing
"exec $::NMWiz::pybin $::NMWiz::prody {*}$args"
    (procedure "::NMWiz::prodySubmitPCAjob" line 61)
    invoked from within
"::NMWiz::prodySubmitPCAjob"
    (procedure "::NMWiz::prodySubmitJob" line 26)
    invoked from within
"::NMWiz::prodySubmitJob"
    invoked from within
".nmwizprody.submitFrame.prodySubmit invoke"
    ("uplevel" body line 1)
    invoked from within
"uplevel #0 [list $w invoke]"
    (procedure "tk::ButtonUp" line 23)
    invoked from within
"tk::ButtonUp .nmwizprody.submitFrame.prodySubmit"
    (command bound to event)

Any clues as to what's going on and how this could be fixed?

dwalgenbach:
I'm having this same problem. I am running it on Windows 10 64 bit .

ZhuoLi1986:
I am facing the same issue, have you figured this out?

ZhuoLi1986:
Thank you very much for the prompt reply. I actually selected a set of residues and CA to do the PCA analysis.  I am not familiar with programming, thus it is hard for me to tell the issue. But I will try to update anything that listed. Thanks a lot.

jamesmkrieger:
I'll look into it and see if I can find the error.

derPuntigamer:
I have the same issue with VMD. I'm not sure if it a VMD related issue or not.

hira-jabeen:
Hi,
 
I am using vmd-1.9.3 and python3.8.0, for PCA analysis through ProDy in VMD.
I am having same issue:

@> Superposing  [ 33%] 1s
                         
@> Superposing  [ 66%] 1s
                         
@> Superposing  [100%] 0s

@> Superposing  [ 33%] 1s
                         
@> Superposing  [ 66%] 1s
                         
@> Superposing  [100%] 0s


[...]

PR #1270, opened 2021-01-20
pca app/nmwiz fix

jamesmkrieger:
Otherwise the NMWiz ProDy Interface throws errors during PCA:
1. The number of modes is more than we have.
2. The output of the LOGGER about superposition.


jamesmkrieger:
closes #746

TianmingQu:
I am sorry, how is this fixed?

jamesmkrieger:
This suppressed the logging so that NMWiz no longer complains like you saw. This should be an old enough change to feature in ProDy 2.0.1 so you should get the fix by updating ProDy

PR #1291, opened 2021-01-29
added dci

jamesmkrieger:
Added dynamic coupling index (DCI; also known as functional dynamic flexibility index [F-DFI]). 

Also, made both DCI and DFI user-facing.


jamesmkrieger:
I think this has been tested enough now.

PR #1292, opened 2021-01-29
New file with protein and protein-ligand interaction is added.

karolamik13:
I created new module which compute protein and protein-ligand interactions.
For protein-ligand interactions there are some additional requirements i.e. installation of Openbabel and PLIP.
There is also new function which after Openbabel installation can provide adding hydrogens to protein and ligand structures.

### DIFF
--- a/prody/proteins/blastpdb.py
+++ b/prody/proteins/blastpdb.py
@@ -264,7 +264,7 @@ class PDBBlastRecord(object):
                             query_len)
                 data['percent_coverage'] = p_overlap
                 
-                for item in (hit['id'] + hit['def']).split('>gi'):
+                for item in (hit['id'] + hit['def']).split('>pdb'):
                     head, title = item.split(None, 1)
                     head = head.split('|')
                     pdb_id = head[-2].lower()


jamesmkrieger:
It could be good to provide an option for adding hydrogens using PDBFixer too.

prody/proteins/interactions.py
```diff
@@ -0,0 +1,1109 @@
...
+    SaltBridges_list = sorted(SaltBridges_list, key=lambda x : x[-1])
+    [ SaltBridges_list.remove(j) for i in SaltBridges_list for j in SaltBridges_list if Counter(i) == Counter(j) ]
+
+    LOGGER.info("Number of detected salt bridges: {0}.".format(len(SaltBridges_list)))        
+
+    return SaltBridges_list
+    
+
+def calcRepulsiveIonicBonding(atoms, distA=4.5):
+    """Searching for repulsive ionic bondning in protein structure
```
jamesmkrieger, line 317: There's a typo here: bondning has an extra n.

prody/proteins/interactions.py
```diff
@@ -0,0 +1,1109 @@
...
+    i.e. between positive-positive or negative-negative residues.
+    Histidine is not considered as a charge residue
```
jamesmkrieger, line 319: charged residue with a d

prody/proteins/interactions.py
```diff
@@ -0,0 +1,1109 @@
...
+                            'with `getCoords` method')
+    try:
+        from plip.structure.preparation import PDBComplex   
+        
+        pdb_name = atoms.getTitle()+'_sele.pdb'
+        LOGGER.info("Writing PDB file with selection in the local direvtory.")
+        writePDB(pdb_name, atoms)
+
+        try:
+            if atoms.hydrogen == None or atoms.hydrogen.numAtoms() < 30: # if there is no hydrigens in PDB structure
```
jamesmkrieger, line 772: typo in comment: should be hydrogens with an o not an i

prody/proteins/interactions.py
```diff
@@ -0,0 +1,1109 @@
...
+                addHydrogens(pdb_name)
+                pdb_name = pdb_name[:-4]+'_addH.pdb'
+                atoms = parsePDB(pdb_name)
+                LOGGER.info("Lack of hydrogen in the structure. Hydrogens will be added.")
```
jamesmkrieger, line 776: in this case you should say have been added or more the logger info up.

prody/proteins/interactions.py
```diff
@@ -0,0 +1,1109 @@
...
+                m1 = resIDs_with_resChIDs.index((int(ii[0][3:]),ii[2]))
+                m2 = resIDs_with_resChIDs.index((int(ii[3][3:]),ii[5]))
+                InteractionsMap[m1][m2] = InteractionsMap[m1][m2] + scoring[nr_i]
+  
+    return InteractionsMap
+
+
+def calcLigandInteractions(atoms, select='all not (water or protein or ion)', ignore_ligs=['NAG','BMA','MAN']):
+    """Provide ligand interactions with other elements of the system including protein, water and ions.
+    Results are computed by PLIP program which should be installed in local computer.
```
jamesmkrieger, line 729: could do with some extra articles/etc here: "the PLIP program" and "your local computer"

prody/proteins/interactions.py
```diff
@@ -0,0 +1,1109 @@
...
+    
+    :arg select: a selection string for residues of interest
+            default is 'all not (water or protein or ion)'
+    :type select: str
+    
+    :arg ignore_ligs: List of ligands which will be excluded from the analysis.
+    :type ignore_ligs: List of str
+    
+    To display results as a list of interactions use showLigandInteractions()
+    and for visualization in VMD program please use showLigandInteraction_VMD()
```
jamesmkrieger, line 743: I'd drop the word program here

prody/proteins/interactions.py
```diff
@@ -0,0 +1,1109 @@
...
+    :type atoms: :class:`.Atomic`
+    
+    :arg select: a selection string for residues of interest
+            default is 'all not (water or protein or ion)'
+    :type select: str
+    
+    :arg ignore_ligs: List of ligands which will be excluded from the analysis.
+    :type ignore_ligs: List of str
+    
+    To display results as a list of interactions use showLigandInteractions()
```
jamesmkrieger, line 742: I'm not sure show is the right starting word for getting a list. It suggests something more visual. You could make this `listLigandInteractions`. This could also be better as a private function that gets called by `calcLigandInteractions` based on a keyword argument such as `output_format`.

prody/proteins/interactions.py
```diff
@@ -0,0 +1,1109 @@
...
+    Pi_stacking = calcPiStacking(atoms.protein)                       #4
+    Pi_cation = calcPiCation(atoms.protein)                           #5
+    Hydroph_calculations = calcHydrophohic(atoms.protein)             #6
+    AllInteractions = [HBs_calculations, SBs_calculations, SameChargeResidues, Pi_stacking, Pi_cation, Hydroph_calculations]   
+    
+    return AllInteractions
+    
+
+def calcInteractionMatrix(atoms):
+    """Calculate matrix with protein interactions which is scored by as follows:
```
jamesmkrieger, line 682: remove by

jamesmkrieger: You should probably explain the origin of these scores/weights and give users the option to change them like with the gap and mismatch scores for sequence alignment.

prody/proteins/interactions.py
```diff
@@ -0,0 +1,1109 @@
...
+                                                     round(RingRing_distance,3), round(RingRing_angle,3)])
+    
+    PiStack_calculations = sorted(PiStack_calculations, key=lambda x : x[-2])   
+    LOGGER.info("Number of detected Pi stacking interactions: {0}.".format(len(PiStack_calculations)))
+    
+    return PiStack_calculations
+
+
+def calcPiCation(atoms, distA=5.0, extraSele=None, **kwargs):
+    """Searching for cation-Pi interaction i.e. between aromatic ring and positively charged residue (ARG and LYS)
```
jamesmkrieger, line 448: I don't think searching is the right form of the verb. You could put search as a command or searches as a description.

karolamik13:
Thank you James.
Best, Karolina

pon., 1 lut 2021 o 13:24 James Krieger <notifications@github.com>
napisał(a):

karolamik13:
All except adding PDBFixer is ready.

jamesmkrieger:
Sounds good

prody/proteins/interactions.py
```diff
@@ -712,7 +715,11 @@ def calcInteractionMatrix(atoms):
     resIDs = list(atoms.select('name CA').getResnums())
     resChIDs = list(atoms.select('name CA').getChids())
     resIDs_with_resChIDs = list(zip(resIDs, resChIDs))
-    scoring = [2, 3, -1, 3, 3, 1]  #HBs_calculations, SBs_calculations, SameChargeResidues, Pi_stacking, Pi_cation, Hydroph_calculations
+
+    try:
+        scoring = kwargs['scoring']
+    except:
+        scoring = [2, 3, -1, 3, 3, 1]  #HBs_calculations, SBs_calculations, SameChargeResidues, Pi_stacking, Pi_cation, Hydroph_calculations
```
jamesmkrieger, line 722: You can set the default values from kwargs.get

also, it might be better to have separate kwargs for each so people don’t have to remember the order and defaults

prody/proteins/interactions.py
```diff
@@ -1098,7 +1114,12 @@ def showInteractionMap(atoms):
             raise TypeError('coords must be an object '
                             'with `getCoords` method')
 
-    InteractionsMap = calcInteractionMatrix(atoms)
+    try:
+        scoring = kwargs['scoring']
+    except:
+        scoring = [2, 3, -1, 3, 3, 1]  #HBs_calculations, SBs_calculations, SameChargeResidues, Pi_stacking, Pi_cation, Hydroph_calculations
```
jamesmkrieger, line 1120: Same comment here

jamesmkrieger (changes requested):
Everything looks good except for kwargs for interaction scores

prody/proteins/interactions.py
jamesmkrieger: I meant articles like “the” and “a”

the journal article is good to have too though

jamesmkrieger: Ok the grammar works now

prody/proteins/interactions.py
jamesmkrieger: This not done yet. It should be “the VMD program” or just “VMD” without program

prody/proteins/interactions.py
jamesmkrieger: Same problem a few lines down in a logger output

jamesmkrieger (commented):
A few more minor things

jamesmkrieger (approved):

prody/proteins/interactions.py
jamesmkrieger: The code would look as follows:
```
scoring = kwargs.get('scoring', [2, 3, -1, 3, 3, 1])  #HBs_calculations, SBs_calculations, SameChargeResidues, Pi_stacking, Pi_cation, Hydroph_calculations
```

prody/proteins/interactions.py
jamesmkrieger: Actually, we don't need the kwargs like this. We want to do it the other way round and provide them straight through to calcInteractionMatrix. 

You could delete these few lines as calcInteractionMatrix has them already and this function doesn't use them itself.

jamesmkrieger (changes requested):
The scoring kwargs thing still needs fixing

jamesmkrieger (approved):

prody/proteins/interactions.py
```diff
@@ -0,0 +1,1209 @@
...
+    for k in pairList:
+        if 180-angle < float(k[-1]) < 180 and float(k[-2]) < distA:
+            aa_donor = atoms.getResnames()[k[0]]+str(atoms.getResnums()[k[0]])
+            aa_donor_atom = atoms.getNames()[k[0]]+'_'+str(k[0])
+            aa_donor_chain = atoms.getChids()[k[0]]
+            aa_acceptor = atoms.getResnames()[k[2]]+str(atoms.getResnums()[k[2]])
+            aa_acceptor_atom = atoms.getNames()[k[2]]+'_'+str(k[2])
+            aa_acceptor_chain = atoms.getChids()[k[2]]
+            
+            HBs_list.append([aa_donor, aa_donor_atom, aa_donor_chain, aa_acceptor, aa_acceptor_atom,
```
jamesmkrieger, line 234: Perhaps you could order it this way round like in the second part of the SHEET header records of the PDB (taken from line 1578 of 2tbv):

`O  LEU A 263   N  VAL A 103`

See http://www.wwpdb.org/documentation/file-format-content/format33/sect5.html#SHEET


This would include splitting resnum and resname, which is probably a good idea anyway.

PR #1312, opened 2021-02-18
typo fix

jamesmkrieger:
deprecation with an e instead of an a


jamesmkrieger:
we should check that it's not needed for anything, e.g. NMWiz, but probably it isn't.

jamesmkrieger:
Ok, I checked and NMWiz isn't using it. None of the rest of ProDy is using it either. It seems to be something that could be useful but hasn't been used. There are DeprecationWarning and PendingDeprecationWarning classes and one is called somewhere but it doesn't depend on the state of the DEPRECATION_WARNINGS global variable. That variable is only defined in this function and doesn't even have a default value otherwise. 

Therefore, it's safe to change it to correct the typo without affecting anything.

PR #1318, opened 2021-02-22
sampleModes docs fix

jamesmkrieger:
This may fix the problem but I'm not sure. The problem is that there is a huge exclamation mark instead of a right bracket.


jamesmkrieger:
Ok. That didn't fix it.

jamesmkrieger:
Actually the page just isn't changing. It didn't respond to the numpy fix either.

jamesmkrieger:
It looks like everything is already fixed now anyway

PR #1350, opened 2021-04-26
emdfile docs

jamesmkrieger:
Here is a first start towards documenting TRN. I also made parseEMD pass kwargs through to it.


jamesmkrieger:
passed checks: https://github.com/jamesmkrieger/ProDy/actions/runs/1615419636

Issue #1364, opened 2021-06-09
parseMMCIF Error reading 1ezc.cif and 2wbc.cif


SHZ66:
The following error was raised while running the ProDy app on `1ezc.cif` and `2wbc.cif`. However, based on the traceback, it looks like this error is not with the app but the mmCIF reader. 

```
Traceback (most recent call last):
  File "/usr/local/bin/prody", line 8, in <module>
    sys.exit(prody_main())
  File "/usr/local/lib/python3.9/site-packages/prody/apps/prody_apps/__init__.py", line 53, in prody_main
    namespace.func(namespace)
  File "/usr/local/lib/python3.9/site-packages/prody/apps/prody_apps/prody_gnm.py", line 300, in <lambda>
    subparser.set_defaults(func=lambda ns: prody_gnm(ns.__dict__.pop('pdb'),
  File "/usr/local/lib/python3.9/site-packages/prody/apps/prody_apps/prody_gnm.py", line 58, in prody_gnm
    pdb = prody.parseMMCIF(pdb, model=model)
  File "/usr/local/lib/python3.9/site-packages/prody/proteins/ciffile.py", line 110, in parseMMCIF
    result = parseMMCIFStream(cif, chain=chain, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/prody/proteins/ciffile.py", line 180, in parseMMCIFStream
    ag = _parseMMCIFLines(ag, lines, model, chain, subset,
  File "/usr/local/lib/python3.9/site-packages/prody/proteins/ciffile.py", line 233, in _parseMMCIFLines
    models.append(line.split()[fields['pdbx_PDB_model_num']])
KeyError: 'pdbx_PDB_model_num'
```

@jamesmkrieger Could you give it a look?

jamesmkrieger:
This is now fixed. Just waiting for checks and then will merge it.

PR #1366, opened 2021-06-10
ciffile fix

jamesmkrieger:
Addresses #1364 

The problem was that line 501 starts with ATOMS as part of a sentence and we hadn't accounted for lines like that.

### DIFF
--- a/prody/proteins/ciffile.py
+++ b/prody/proteins/ciffile.py
@@ -226,7 +226,7 @@ def _parseMMCIFLines(atomgroup, lines, model, chain, subset,
             fieldCounter += 1
             fields[line.split('.')[1].strip()] = fieldCounter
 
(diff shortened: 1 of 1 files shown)


prody/proteins/ciffile.py
```diff
@@ -226,7 +226,7 @@ def _parseMMCIFLines(atomgroup, lines, model, chain, subset,
             fieldCounter += 1
             fields[line.split('.')[1].strip()] = fieldCounter
 
-        if line.startswith('ATOM') or line.startswith('HETATM'):
+        if line.startswith('ATOM  ') or line.startswith('HETATM'):
```
SHZ66, line 229: Should it be ATOMS or ATOM here? just wondering

jamesmkrieger: Should be ATOM with two spaces or HETATM. These are the two options.

PR #1378, opened 2021-06-30
Transformation getMatrix fix

jamesmkrieger:
Otherwise, this if condition sometimes raises an error as None has no attribute _matrix


jamesmkrieger:
Actually this needs revisiting as it doesn't make sense. self can't be None or it wouldn't have all those methods.

prody/measure/transform.py
```diff
@@ -71,7 +71,7 @@ def getMatrix(self):
         """Returns a copy of the 4x4 transformation matrix whose top left is
         rotation matrix and last column is translation vector."""
 
-        if self._matrix is not None:
+        if self is not None and self._matrix is not None:
```
SHZ66, line 74: I don't think this line would be ever reached if `self` were somehow `None`, because `None` wouldn't be an instance of this class

SHZ66 (changes requested):

prody/measure/transform.py
jamesmkrieger: yeah. I’ll have to look into it again

PR #1383, opened 2021-07-15
Vector crosscorr

jamesmkrieger:
This allows Vector objects such as deformation vectors to be used for calculating NxN cross-correlation matrices. It already works for 3Nx3N covariance matrices. 

Also, clarified docs a bit.


jamesmkrieger:
passed checks: https://github.com/jamesmkrieger/ProDy/actions/runs/1601646053

PR #1384, opened 2021-07-15
added showCovariance for 3Nx3N

jamesmkrieger:
We didn't have this so I added it


jamesmkrieger:
passes checks: https://github.com/jamesmkrieger/ProDy/actions/runs/1601618552

PR #1390, opened 2021-07-29
showAtomicMatrix asym atoms click/zoom fix

jamesmkrieger:
The interactive parts of showAtomicMatrix could not previously handle asymmetric atoms for x and y axes e.g. 

```
showAtomicMatrix(distance_matrix[:9, 9:17], atoms=(resis[:9], resis[9:17]))
```

Clicking or zooming would yield the following error, and the clicked point would not be labelled with residues.

```
Traceback (most recent call last):
  File "C:\Users\james\.conda\envs\prody\lib\site-packages\matplotlib\cbook\__init__.py", line 224, in process
    func(*args, **kwargs)
  File "C:\Users\james\code\ProDy\prody\utilities\misctools.py", line 72, in onClick
    self.show(event)
  File "C:\Users\james\code\ProDy\prody\utilities\misctools.py", line 124, in show
    seq = self.atoms.getSequence()
AttributeError: 'tuple' object has no attribute 'getSequence'
```

With this fix, there is no longer such a problem.


jamesmkrieger:
passes checks: https://github.com/jamesmkrieger/ProDy/actions/runs/1601627226

PR #1395, opened 2021-08-01
parsePDB_bonds_fix

### LINKED ISSUE #1385: Off by one when reading connections from pdb?
There appear to be a few issues when reading bonds from PDB files.

In the function _parsePDBLines, In this situation

```
        elif not onlycoords and (startswith == 'TER   ' or
            startswith.strip() == 'TER'):
            termini[acount - 1] = True
            acount += 1   # This line is necessary
```
The lack of the increment causes descrepancies in the Atom index in the AtomGroup and the Atom index in the PDB (this is because TER entries increment the atom field)

Secondly, The mechanics of the other parts of the code seems to suggest that the Atom index is 0 starting, and it is this index that is used with the bonds. This means that in the preceding lines

```
        elif startswith == 'CONECT':
            if bonds is not None:
                atom_serial = line[6:11]
                bonded1_serial = line[11:16]
                bonds.append([int(atom_serial)-1, int(bonded1_serial)-1])  # -1 necessary
                
                bonded2_serial = line[16:21]
                if len(bonded2_serial.strip()):
                    bonds.append([int(atom_serial)-1, int(bonded2_serial)-1])  # -1 necessary

                bonded3_serial = line[21:26]
                if len(bonded3_serial.strip()):
                    bonds.append([int(atom_serial)-1, int(bonded3_serial)-1])  # -1 necessary
                    
                bonded4_serial = line[27:31]
                if len(bonded4_serial.strip()):
                    bonds.append([int(atom_serial)-1, int(bonded4_serial)-1])  # -1 necessary
```

the -1 is necessary. Otherwise causing several off-by-one errors.

jamesmkrieger:
fixes #1385

### DIFF
--- a/prody/proteins/pdbfile.py
+++ b/prody/proteins/pdbfile.py
@@ -737,23 +737,24 @@ def _parsePDBLines(atomgroup, lines, split, model, chain, subset,
             if bonds is not None:
                 atom_serial = line[6:11]
                 bonded1_serial = line[11:16]
-                bonds.append([int(atom_serial), int(bonded1_serial)])
+                bonds.append([int(atom_serial)-1, int(bonded1_serial)-1])
                 
                 bonded2_serial = line[16:21]
                 if len(bonded2_serial.strip()):
-                    bonds.append([int(atom_serial), int(bonded2_serial)])
+                    bonds.append([int(atom_serial)-1, int(bonded2_serial)-1])
 
                 bonded3_serial = line[21:26]
                 if len(bonded3_serial.strip()):
-                    bonds.append([int(atom_serial), int(bonded3_serial)])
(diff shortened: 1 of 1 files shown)


jamesmkrieger:
This currently fails the checks (see https://github.com/jamesmkrieger/ProDy/actions/runs/1087744313). I think the problem is that the checks don't take this problem into account and need fixing though.

jamesmkrieger:
Actually, it looks there is a problem with this code too and not necessarily with the tests. Both need looking at carefully.

jamesmkrieger:
This now passes the tests: https://github.com/jamesmkrieger/ProDy/actions/runs/1094250226

PR #1403, opened 2021-08-25
added non-absolute option to showOverlapTable

jamesmkrieger:
This allows us to get signed values (as below) rather than just absolute ones, but we still get absolute ones by default.

sliced_vs_redone_D614G_pca


jamesmkrieger:
The above result is obtained using abs=False.

By default, we still get the following as before:
sliced_vs_redone_D614G_pca_default_abs

PR #1405, opened 2021-09-12
addCoordset and trimPDBEnsemble keep transformations

jamesmkrieger:
Without this, the trimmed ensemble doesn't have transformations set. With it, it does and can be used with alignByEnsemble.


jamesmkrieger:
Passed checks: https://github.com/jamesmkrieger/ProDy/actions/runs/1601578874

PR #1406, opened 2021-09-14
added gromos rmsd clustering

jamesmkrieger:
This method is the same as the one used in https://manual.gromacs.org/documentation/current/onlinehelp/gmx-cluster.html


jamesmkrieger:
Perhaps we should give this a better name

prody/utilities/catchall.py
```diff
@@ -968,3 +968,51 @@ def findSubgroups(tree, c, method='naive', **kwargs):
...
+
+def findRMSDClusters(rmsd_matrix, c, labels=None):
+    """
+    Divide **rmsd_matrix** into clusters using the gromos method and a cutoff **c**.
+    Returns a list of lists with labels divided into clusters.
+    """    
+    clusters = []
+
+    useful_rmsd_matrix = rmsd_matrix
+    elements = labels
```
jamesmkrieger, line 984: Do we need a variable called elements rather than just using labels?

prody/utilities/catchall.py
```diff
@@ -968,3 +968,51 @@ def findSubgroups(tree, c, method='naive', **kwargs):
...
+    
+    indices = list(range(len(elements)))
+    if labels is None:
+        labels = indices
```
jamesmkrieger, line 988: how can this work if labels is none and indices comes from elements, which comes from labels?

jamesmkrieger (commented):
Definitely, something isn't right here for the labels=None case. Perhaps make labels a necessary argument rather than defaulting to None

jamesmkrieger:
> Definitely, something isn't right here for the labels=None case. Perhaps make labels a necessary argument rather than defaulting to None

Actually, we do want to support the case where we have no labels, e.g. from an MD trajectory. It was just done wrong.

jamesmkrieger:
It should be good now, but needs testing

jamesmkrieger:
This works to cluster a set of trajectories now:
gromos_clusters_2p2A

jamesmkrieger:
```
from prody import *
plt.ion()
import numpy as np

dirname = 'prody_pca/'
ref_pca = loadModel(dirname+'ref.pca.npz')

rmsd_matrix = ens.getRMSDs(pairwise=True)

x = calcRMSDclusters(rmsd_matrix, 2.1)

colors = np.zeros(num_confs, dtype=int)
for i, c in enumerate(x):
    colors[list(np.array(c, dtype=int))] = i
colors = list(colors)

plt.figure()
showProjection(ens, ref_pca[:3], markersize=3, color=colors)
plt.savefig('gromos_clusters_2p1A.png')
```

PR #1418, opened 2021-10-25
select deprec fix

### LINKED ISSUE #1417: module 'pyparsing' has no attribute 'operatorPrecedence'
File "prody/atomic/select.py", line 1090, in _getParser
    parser = pp.operatorPrecedence(expr, oplist)
AttributeError: module 'pyparsing' has no attribute 'operatorPrecedence'

Upon investigation, it's because 'operatorPrecedence' is deprecated in favour of 'infixNotation'. Not sure which versions should match up, but installing ProDy installs the latest version of pyparsing, which doesn't have that function anymore.

Changing the call to infixNotation instead seems to work.

jamesmkrieger:
fixes #1417

### DIFF
--- a/prody/atomic/select.py
+++ b/prody/atomic/select.py
@@ -449,6 +449,10 @@ from numpy import invert, unique, concatenate, all, any
 from numpy import logical_and, logical_or, floor, ceil, where
 
 import pyparsing as pp
+try:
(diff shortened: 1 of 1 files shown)


jamesmkrieger:
passed checks: https://github.com/jamesmkrieger/ProDy/actions/runs/1408630369

PR #1433, opened 2021-12-16
added calcRWSIP function and calcRMSIP and calcSIP aliases

jamesmkrieger:
- adds calcRWSIP for root weighted square inner product from Carnevale et al. 2007
- adds calcRMSIP, which is an alias for calcSubspaceOverlap
- fixes reference formatting on calcSquareInnerProduct
- adds calcSIP, which is an alias for calcSquareInnerProduct


jamesmkrieger:
This actually doesn't work yet

jamesmkrieger:
Forgot to put a sum in the denominator and put eigvecs instead of eigvals before. 

This still isn't quite right though as my value is much lower than expected (0.33 vs 0.74).

jamesmkrieger:
> This still isn't quite right though as my value is much lower than expected (0.33 vs 0.74).

This is because, the eigenvalues to be used should be 1/variance when using PCA. Now  have 0.745

jamesmkrieger:
We still need tests for these functions and dynamics/compare functions generally

PR #1439, opened 2021-12-21
Iterable import fix for python 3.10

jamesmkrieger:
This accounts for changes in python 3.10 as noted in https://github.com/conda-forge/prody-feedstock/pull/10

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [2.7, 3.5, 3.9]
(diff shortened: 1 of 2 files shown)


jamesmkrieger:
This doesn't work: https://github.com/jamesmkrieger/ProDy/actions/runs/1608093432

Looks like Biopython doesn't support Python 3.10 yet

jamesmkrieger:
This is now redundant with #1440 

Still would be good to double check that no other imports need fixing too.

jamesmkrieger:
This failed the checks anyway as Biopython and SciPy are not yet compatible with Python 3.10: https://github.com/jamesmkrieger/ProDy/runs/4777405586?check_suite_focus=true

PR #1455, opened 2022-01-25
Fixed a bug where weights are indexed twice

SHZ66:
The weights don't need to be indexed in Conformation because they are already in Ensemble.


jamesmkrieger (approved):
Looks good and explains the bug we had in a much more sensible way. Thanks

PR #1456, opened 2022-01-26
Reorder imports to avoid deprecation warning

kalekundert:
In `atomic/select.py`, importing `Iterable` from `collections` instead of `collections.abc` causes a `DeprecationWarning` in modern versions of python.  This commit avoids the warning by first trying to import from `collections.abc` before falling back on `collections`.

It'd probably be fine to remove the `collections` import altogether, since `collections.abc` has been present since python 3.3.  But in the interest of being conservative, I just changed the order in which the imports are tried.

### DIFF
--- a/prody/atomic/select.py
+++ b/prody/atomic/select.py
@@ -442,11 +442,11 @@ all alphanumeric characters."""
 import sys
 from re import compile as re_compile
 try:
-   # for upto python 3.9
-   from collections import Iterable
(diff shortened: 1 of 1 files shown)


jamesmkrieger:
Ok, I guess that makes sense to avoid the warning. Thanks

We do want to keep compatibility with Python 2.7 for building the website

PR #1460, opened 2022-01-31
Clustenm reversions

jamesmkrieger:
takes into account good parts of #1459 but makes it more stable and reverts docs changes that don't match elsewhere


jamesmkrieger:
Now this pull request only has better import errors to say about needing OpenMM 7.6

jamesmkrieger:
Thanks and you're welcome

PR #1474, opened 2022-02-17
Final release changes

jamesmkrieger:
I may update this more or this may be it. We'll see.


jamesmkrieger:
Just a confirmation that this passes the checks: https://github.com/jamesmkrieger/ProDy/actions/runs/1860473437

We still need to work out how to trigger them automatically though

PR #1480, opened 2022-03-01
cealign non-ca fix

jamesmkrieger:
Without this fix, using CE alignment on all-atom structures is prohibitively slow. Taking only Calpha coords fixes it.

### DIFF
--- a/prody/proteins/compare.py
+++ b/prody/proteins/compare.py
@@ -1394,8 +1394,13 @@ def getCEAlignMapping(target, chain):
                     'It may not be installed properly.')
         return None
 
-    tar_coords = target.getCoords(calpha=True).tolist()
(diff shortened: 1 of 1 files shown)


jamesmkrieger:
Actually, this doesn't work yet as the resulting atom maps have completely wrong selection strings.

SHZ66:
What would happen if we align two full-atom structures with this modification? Would the resulting atommap only have calpha carbons? If so, I'd say we should let the user explicitly specify CA when they align structures.

jamesmkrieger:
No, there’s a step to extend the mapping to the rest of the residue

SHZ66:
Ok, in that case `calpha=True` would make sense. It sounds good to have a check for protein too.

jamesmkrieger:
Ok, thanks

PR #1499, opened 2022-03-13
modehunter interpolateModel

jamesmkrieger:
This function for thin plate spline interpolation has been copied and modified from the Python 2 code in SITUS ModeHunter in accordance with its legal notice. 

It should be a very useful functionality for linking coarse-grained and pseudoatom modes to atomic detail.


jamesmkrieger:
I guess this should be tested with something besides Calpha and all atoms to make sure it really works as expected. 

For going from Calpha to all atoms, it definitely does work as we get reasonable overlaps (maybe 0.7 or so for most modes) to the output of extendModel.

jamesmkrieger:
We're looking into thin plate splines interpolation more and will see if we can get this working better

jamesmkrieger:
I think the answer is that we can't and this should just go in as is after one more check through

PR #1515, opened 2022-03-29
Extend ClustENM kwargs for more functionality

jamesmkrieger:
Here are some suggested improvement for ClustENM that are helpful for Scipion. 

- Add more kwargs for ANM options to include sparse and kdtree for buildHessian and turbo for calcModes
- Add threads kwarg to control number of threads per simulation. This requires the CPU platform and that has been documented.
- Extend parallel kwarg to allow the user to set the number of parallel simulations to be run at a time. Setting just 1 simulation in parallel sets parallel to False and setting 0 sets it to True to use all CPUs as previously. Original True and False options are still accepted too.

### DIFF
--- a/prody/dynamics/clustenm.py
+++ b/prody/dynamics/clustenm.py
@@ -478,7 +478,8 @@ class ClustENM(Ensemble):
     def _buildANM(self, cg):
 
         anm = ANM()
-        anm.buildHessian(cg, cutoff=self._cutoff, gamma=self._gamma)
+        anm.buildHessian(cg, cutoff=self._cutoff, gamma=self._gamma,
+                         sparse=self._sparse, kdtree=self._kdtree)
 
         return anm
 
@@ -513,7 +514,7 @@ class ClustENM(Ensemble):
         if not self._checkANM(anm_cg):
             return None
 
-        anm_cg.calcModes(self._n_modes)
+        anm_cg.calcModes(self._n_modes, turbo=self._turbo)
 
         anm_ex = self._extendModel(anm_cg, cg, tmp)
         ens_ex = sampleModes(anm_ex, atoms=tmp,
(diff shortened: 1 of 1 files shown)


jamesmkrieger:
Ok, I can call it something besides threads. It does seem to engage a certain number of cpus and spread the work across them. You’re right I don’t know what the right word to use is. Does processes or cpus work?

I am aware that there are separate parts for conformer generation and simulation and only the cluster generation could be in parallel. Nevertheless, I did manage to do something to get the simulations to use a particular number of cpus too. I think I was dividing the number of cpus requested by the number of cluster representatives. 

I didn’t do any benchmarking and definitely not on large systems. I just did a few small tests on AK. Even there, I’d agree the CPU platform of OpenMM wasn’t great. 

The idea wasn’t to get good efficiency though but rather to free up some of the computer for other things. 

I guess an alternative way to do this could be to wrap the whole clustenm execution with threadpoolctl or something to control the number of cpus used from the outside as I did for the ANM app in https://github.com/prody/ProDy/pull/1516. I actually found that using almost all cpus was faster than using all of them like the default for a few systems including fairly big ones like the AMPAR and spike. 

I just thought it might be better to make use of what ClustENMD and OpenMM already have, but perhaps not.

jamesmkrieger:
What I’m trying to add in my Scipion plugin is in https://github.com/scipion-em/scipion-em-prody/pull/13

prody/dynamics/clustenm.py
```diff
@@ -967,10 +984,17 @@ def run(self, cutoff=15., n_modes=3, gamma=1., n_confs=50, rmsd=1.0,
         :arg platform: Architecture on which the OpenMM part runs, default is None.
             It can be chosen as 'CUDA', 'OpenCL' or 'CPU'.
             For efficiency, 'CUDA' or 'OpenCL' is recommended.
+            'CPU' is needed for setting threads per simulation.
         :type platform: str
 
         :arg parallel: If it is True (default is False), conformer generation will be parallelized.
+            This can also be set to a number for how many simulations are run in parallel.
```
jamesmkrieger, line 991: This line is flawed I agree. It should say how many parallel processes are used for pooling the conformer generation and not have anything to do with simulations.

prody/dynamics/clustenm.py
```diff
@@ -967,10 +984,17 @@ def run(self, cutoff=15., n_modes=3, gamma=1., n_confs=50, rmsd=1.0,
...
+            Setting 0 or True means run as many as there are CPUs on the machine.
         :type parallel: bool
+
+        :arg threads: Number of threads to use for an individual simulation
```
jamesmkrieger, line 995: This one that I call threads though is correct. This is linking up to the threads setting in OpenMM

jamesmkrieger:
Actually, I am using the threads argument for OpenMM threads so that part isn't a problem. 

I am then separately setting a number of parallel CPUs for the conformer generation and yes, you're right I documented it wrong.

I've now clarified these things in revised docs. 

No idea why the Python 3.5 check failed. It seems to have not managed to start properly

PR #1516, opened 2022-03-29
anm app accept sparse kdtree turbo

jamesmkrieger:
I made sure these use available letters.

h was available as hessian writing uses l to go with k for kirchhoff


jamesmkrieger:
This has now been more carefully checked with from prody.apps import prody_parser

jamesmkrieger:
The number of processors doesn't make much difference for PCA as it's done so quickly anyway, although this may change for large simulations. For ANM and GNM, it's clearly working

PR #1535, opened 2022-04-25
Fix typo in 'calcAnisousFromModel()'

padix-key:
The docstring of 'calcAnisousFromModel()' states that a 3Nx6 matrix is returned, but in reality a Nx6 matrix is returned, since there a 6 different `U` values for each atom.

### DIFF
--- a/prody/dynamics/analysis.py
+++ b/prody/dynamics/analysis.py
@@ -677,7 +677,7 @@ def calcHitTime(model, method='standard'):
 
 
 def calcAnisousFromModel(model, ):
-    """Returns a 3Nx6 matrix containing anisotropic B factors (ANISOU lines)
+    """Returns a Nx6 matrix containing anisotropic B factors (ANISOU lines)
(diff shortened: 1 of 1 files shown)


prody/dynamics/analysis.py
```diff
@@ -747,4 +747,4 @@ def calcScipionScore(modes):
 
     return score
 
-calcHemnmaScore = calcScipionScore
\ No newline at end of file
+calcHemnmaScore = calcScipionScore
```
jamesmkrieger, line 750: what happened here?

jamesmkrieger:
Thanks for spotting this

prody/dynamics/analysis.py
padix-key: *GitHub* seems to have added an empty line at the end of file. I created this PR directly via the editor on *GitHub*.

jamesmkrieger: ok, I think that's reasonable

PR #1536, opened 2022-04-26
Fixing bond (CONECT) loading within pdb file parsing

zrqiao:
# Description
I am recently working on loading the atom connectivity information from pdb files. While parsing files with `parsePDB(file, bond=True)`, I noticed that for several files with missing/unresolved atoms the connectivities are incorrectly assigned to the atomic indices of the `AtomGroup` object. 

Taking further inspections, I noticed that in `prody.proteins.pdbfile._parsePDBLines()` the serial numbers loaded from the CONECT entry are not re-matched to the atom indices. I have created a tentative fix that works correctly for files that I have tested on. A typo in line parsing is corrected alongside.

## Type of change
- [x] Bug fix

I am happy to discuss a bit and perform minor code refactoring to get it merged. As a side note, I wonder if prody fully supports loading bond information from mmCIF files?

### DIFF
--- a/prody/atomic/atomgroup.py
+++ b/prody/atomic/atomgroup.py
@@ -1095,8 +1095,8 @@ class AtomGroup(Atomic):
         index 1"``.  See :mod:`.select` module documentation for details.
         Also, a data array with number of bonds will be generated and stored
         with label *numbonds*.  This can be used in atom selections, e.g.
-        ``'numbonds 0'`` can be used to select ions in a system. If *bonds*
-        is empty or **None**, then all bonds will be removed for this
+        ``'numbonds 0'`` can be used to select ions in a system. If *bonds* 
+        is empty or **None**, then all bonds will be removed for this 
         :class:`.AtomGroup`. """
 
         if bonds is None or len(bonds) == 0:
@@ -1126,7 +1126,7 @@ class AtomGroup(Atomic):
         self._fragments = None
 
     def numBonds(self):
-        """Returns number of bonds.  Use :meth:`setBonds` or
+        """Returns number of bonds.  Use :meth:`setBonds` or 
         :meth:`inferBonds` for setting bonds."""
 
         if self._bonds is not None:
@@ -1134,7 +1134,7 @@ class AtomGroup(Atomic):
         return 0
 
     def getBonds(self):
-        """Returns bonds.  Use :meth:`setBonds` or
+        """Returns bonds.  Use :meth:`setBonds` or 
         :meth:`inferBonds` for setting bonds."""
 
         if self._bonds is not None:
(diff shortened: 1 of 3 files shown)


jamesmkrieger:
I don't think it does support loading of bonding information from mmCIF files yet. Do you have an example we can look at?

jamesmkrieger:
I'm not sure about this fix. It looks like you've tried to fix what was already an attempt to fix this, which is good if it was broken. I'll have to look at it more carefully another time.

Also, I'd suggest not making all these style fixes because it makes it much harder to find the meaningful changes.

zrqiao:
@jamesmkrieger Can you please point me to the previous fix attempt? I was not aware of another branch of PR. The current code misassign bonds for cases where the serial numbers are not consecutive, e.g., 6XMK in pdb.

Sorry about the confusion on style changes, these seem to be added by my editor automatically. I will remove that.

jamesmkrieger:
I made a fix at https://github.com/prody/ProDy/pull/1395 because the numbers associated with bonds were drifting off because of not counting TER records properly. You can see the documentation of the issue and how I fixed it at https://github.com/prody/ProDy/issues/1385.

jamesmkrieger:
I can definitely look at the 6XMK cif file too and see if we can parse the info from there. It seems from #1385 like 3h5v is a good example without your problem to look at too. 

Have you made sure your fix works for pdb files without the serials jumping? You may want to test it on 3h5v too.

zrqiao:
@jamesmkrieger Thanks! Just went through the previous PR, looks like it assumes no serial jump which is failing in my test cases. For 6XMK, I have been using the .pdb file format but not sure if the issue is present in mmCIF/ whether mmCIF bonds can be parsed natively.

zrqiao:
@jamesmkrieger Thanks! Just went through the previous PR, looks like it assumes no serial jump which is failing in my test cases. For 6XMK, I have been using the .pdb file format but not sure if the issue is present in mmCIF/ whether mmCIF bonds can be parsed natively.

jamesmkrieger:
At the moment mmCIF bonds probably can't be parsed, but I don't remember either. Can you check please?

zrqiao:
> I can definitely look at the 6XMK cif file too and see if we can parse the info from there. It seems from #1385 like 3h5v is a good example without your problem to look at too.
> 
> Have you made sure your fix works for pdb files without the serials jumping? You may want to test it on 3h5v too.

I have just tested on 3h5v and look it correctly assigned the bonds; also hand-checked that the ASN-NAG bonds are located correctly. I have reverted some automatic style changes but am having trouble re-adding the leading tabs, hopefully the commits are easier to parse now.

I tested on `1bos.cif` and the current code does not parse the `_struct_conn` section. I am inclined to output a meaningful error/warning when parsing cifs with bonding information.

jamesmkrieger:
Yes, a warning for cifs could be good.

Ok, then I guess your fix is doing something similar to my fix but better. 

Just one thing: have you looked at serials vs indices as things that already exist in ProDy and whether you really need to make your own converter? I guess yours coming from ranges are doing something different, but it would be good to be sure.

jamesmkrieger:
We usually use “{0}”.format(i) and suchlike. You can see some examples earlier in the file

zrqiao:
@jamesmkrieger I have cleaned up the code a bit and added a warning clause. Please let me know if it is fine for merging.

jamesmkrieger:
Yes,  think it looks fine. Thanks again for the fix

PR #1539, opened 2022-04-27
Rmsf

RicardoSerr:
Added functions in order to calculate and plot the RMSF from modes


RicardoSerr:
Based on the showSQFlucts implementation

PR #1540, opened 2022-04-27
ShowAtomicLines is fixed for multiple chains

RicardoSerr:
Fixed overlay options for multiple chains.


RicardoSerr:
Using pdb structure 3kg2, GNM non zero mode 2 should be showing 4 chains lines like in Figure 8 from paper
https://www.cell.com/structure/fulltext/S0969-2126(15)00282-8?_returnURL=https%3A%2F%2Flinkinghub.elsevier.com%2Fretrieve%2Fpii%2FS0969212615002828%3Fshowall%3Dtrue

Original behaviour: the second chain is replaced by later chains
Chains_bad

Fixed behaviour: 4 chains are shown as expected
Chains_OK

PR #1545, opened 2022-05-10
clustenm accept sparse kdtree turbo

jamesmkrieger:
This is the simple, non-objectionable part of #1515


kaynakb:
You're welcome, James. Thanks for the addition.

PR #1549, opened 2022-05-13
WIP: Motif search

mkonstanty:
This feature adds possibility to search for a MOTIF using Prody package.


prody/sequence/__init__.py
```diff
@@ -46,27 +46,41 @@
   * :func:`.showShannonEntropy` - plot Shannon entropy
   * :func:`.showMSAOccupancy` - plot row (sequence) or column occupancy
   * :func:`.showMutinfoMatrix` - show mutual information matrix
+
+
+Searching
+========
+  * :func:`.getPdbCode` - get PDB code from MOTIF
```
jamesmkrieger, line 53: This isn't an informative enough name

prody/sequence/motif.py
```diff
@@ -0,0 +1,125 @@
...
+    r"(\[[ARDNCEQGHILKMFPSTWYV]+\])*"
+    r"({{[{{ARDNCEQGHILKMFPSTWYV}}]+}})*"
+    r"([ARDNCEQGHILKMFPSTWYVx]\(\d(,\d)?\))*)+$"
+)
+MOTIF_MATCHER = re.compile(MOTIF_PATTERN)
+
+
+def getPdbFromMotif(motif: str, database: str) -> str:
+    if database not in DATABASES:
+        raise ValueError(f"Database must be one of: {DATABASES.keys()}.")
```
jamesmkrieger: Please don't use f strings. They don't work with older python versions.

docs/conf.py
```diff
@@ -2,57 +2,62 @@
 import os
 import sys
 
-
-sys.path.append(os.path.abspath('sphinxext'))
```
jamesmkrieger, line 6: Please don't make stylistic changes like this just because. It makes it harder for people to follow the real changes.

prody/database/__init__.py
```diff
@@ -70,31 +70,37 @@
...
+  * :func:`.getLocalRelease` - checks local Swiss-Prot release version
+  * :func:`.checkForUpdates` - checks wheather there is newer Swiss-Prot version than current local one
 
+.. _GOA: https://www.ebi.ac.uk/GOA/
 """
 
-__all__ = []
 
-from . import pfam
+from . import cath, dali, goa, pfam, quartataweb, swissprot, uniprot
```
jamesmkrieger, line 90: please don't reformat like this. It's helpful to be able to see the package imports in blocks like we had

prody/utilities/__init__.py
```diff
@@ -61,19 +61,19 @@
 
 """
 
-__all__ = []
-
+from . import catchall
```
jamesmkrieger, line 64: again, reformatting not appreciated

jamesmkrieger (changes requested):
Thanks for adding this new functionality. It sounds good. I still need to look through it properly though.

In the meantime, if you could reverse all the unrelated formatting changes, that would be make it much easier to follow. Thanks

prody/utilities/helpers.py
```diff
@@ -0,0 +1,53 @@
+# -*- coding: utf-8 -*-
+"""This module defines functions and constants related to MOTIF."""
```
jamesmkrieger, line 2: perhaps this file could be named something more specific

prody/database/refseq.py
```diff
@@ -0,0 +1,160 @@
...
+                    result += line + "\n"
+            return result
+
+    @classmethod
+    def saveInstalledFiles(cls) -> None:
+        """Saves installed files list with check sums locally."""
+        installed_files = cls.getInstalledFiles()
+        path = os.path.join(PRODY_DATA, cls.__name__)
+        os.makedirs(path, exist_ok=True)
+        with open(f"{PRODY_DATA}/{cls.__name__}/{cls.CHECK_SUMS}", "w", encoding="utf-8") as file:
```
jamesmkrieger, line 70: again please don't use an f string

prody/database/refseq.py
```diff
@@ -0,0 +1,160 @@
...
+            file.write(installed_files)
+        LOGGER.debug(f"{cls.CHECK_SUMS} file saved locally.")
```
jamesmkrieger, line 72: here too

prody/sequence/__init__.py
jamesmkrieger: This is now called getPdbFromMotif

jamesmkrieger:
Everything look much better now. Thanks.

Still if you could replace f strings with older types of formatting, that would be great. We need compatibility with python 2.7 for building the prody website.

prody/sequence/motif.py
mkonstanty: > Everything look much better now. Thanks.
> 
> Still if you could replace f strings with older types of formatting, that would be great. We need compatibility with python 2.7 for building the prody website.

hello James,
python 2.7 and 3.5 are end of life, could you please consider abaddon backward compatibility for those versions? - it's not just about f-strings.

jamesmkrieger: No, unfortunately we can't. We can drop 3.5 for sure but not 2.7

If you (or anyone else) know how to build websites with sphinx and can help reconfigure ours then we might be able to but otherwise no.

jamesmkrieger (approved):

prody/sequence/motif.py
```diff
@@ -0,0 +1,491 @@
+# -*- coding: utf-8 -*-
+
+"""This module prase protein sequences using PROSITE motifs."""
```
jamesmkrieger, line 3: I think there's a typo here. `prase` is not a word that I know. Do you mean `parse`? If so, it should be conjugated as `parses`

prody/database/refseq.py
```diff
@@ -0,0 +1,180 @@
+# -*- coding: utf-8 -*-
+
+"""RefSeq database operations."""
+
+import os
+import re
+from concurrent.futures import ThreadPoolExecutor
+
+import requests
+import requests_cache
```
jamesmkrieger, line 10: it's best to make these changes inside your function to not make importing prody fail if someone doesn't install this module.

jamesmkrieger (changes requested):

prody/utilities/motifhelpers.py
```diff
@@ -0,0 +1,166 @@
...
+    with open(filename, "rb") as unknown_file:
+        file_start = unknown_file.read(max_len)
+    compression_type = None
+    for file_type, magic_bytes in magic_dict.items():
+        if file_start.startswith(magic_bytes):
+            compression_type = file_type
+    return compression_type
+
+
+def openFile(filename: str) -> Any:
```
jamesmkrieger, line 112: This takes precedence over the existing openFile in prody/utilities/pathtools because of the import order and breaks other code because it doesn't support other arguments like the original does. Rename or remove it

jamesmkrieger (changes requested):

PR #1551, opened 2022-05-13
no svd in pca app to allow nmodes control

jamesmkrieger:
Using svd meant that the resulting npz file ends up with too many modes in many cases. It's not possible to index it before saveModel.


jamesmkrieger:
It also requires more than 3 coordinate sets according to some ValueError raises it has.

PR #1561, opened 2022-06-07
fix mode ens for empty labels in addModeSets

jamesmkrieger:
Labels are needed for various downstream functions including getEigvecs and getArray so they need to be set no matter what. The if conditions we had inside the main if condition were already enough I think without it. They handle the two cases of no labels in the original object and no labels in the input to the function.


jamesmkrieger:
@SHZ66, was there a reason for not setting them like this? Is it better to try and change the methods that need labels instead?

PR #1563, opened 2022-06-23
Default to https

dkoes:
Otherwise won't work when called from webpages (e.g. colab) that default
to https since you can't make an insecure connection from a secure
webpage.


jamesmkrieger:
Thanks

PR #1588, opened 2022-09-12
Fix IndexError exception when local pairwise alignment produces 0 alignments

ArturKlauser:
It is possible for local alignment between pairs of chains to produce an empty list of alignments, resulting in an IndexError in the current code. This patch takes care of this issue.

Original error:

> File .../prody/proteins/compare.py:796, in getAlignedMatch(ach, bch)
>     789 else:
>     790     alignment = pairwise2.align.globalms(ach.getSequence(),
>     791                                          bch.getSequence(),
>     792                                          MATCH_SCORE, MISMATCH_SCORE,
>     793                                          GAP_PENALTY, GAP_EXT_PENALTY,
>     794                                          one_alignment_only=1)
> --> 796 this = alignment[0][0]
>     797 that = alignment[0][1]
>     798 amatch = []
> 
> IndexError: list index out of range
>

### DIFF
--- a/prody/proteins/compare.py
+++ b/prody/proteins/compare.py
@@ -799,8 +799,9 @@ def getAlignedMatch(ach, bch):
     try:
         this = alignment[0][0]
         that = alignment[0][1]
-    except IndexError:                                                                                                                                                                                                                     
-        return amatch, bmatch, match    
+    except IndexError:
+        LOGGER.warning('Matching chains resulted in empty alignment.')
+        return amatch, bmatch, match
     aiter = ach.__iter__()
     biter = bch.__iter__()
     for i in range(len(this)):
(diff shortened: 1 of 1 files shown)


jamesmkrieger:
Thanks

jamesmkrieger:
Why do you return if you get an index error?

ArturKlauser:
I return an empty alignment in that case. Are there other reasonable options?

jamesmkrieger:
It depends on what upstream functions are expecting. It looks like the only one at the moment is matchChains and that would handle it ok. You'd get 0 seqid and 0 coverage. 

One alternative would be to raise an error, but I guess that's not actually a reasonable option.

Probably, the best thing to do is to keep what you have but raise a warning with LOGGER.warning

ArturKlauser:
Warning sounds good. I updated the PR accordingly.

jamesmkrieger:
Thanks

Issue #1587, opened 2022-09-11
Powering down the Pfam website


boneta:
It seems that the Pfam web site is closing down soon.

>On October 5th, we will start redirecting the traffic from Pfam ([pfam.xfam.org](https://pfam.xfam.org/)) to InterPro ([www.ebi.ac.uk/interpro](https://www.ebi.ac.uk/interpro)). The Pfam website will be available at legacy.pfam.xfam.org until January 2023, when it will be decommissioned. You can read more about the sunset period in our [blog post](https://xfam.wordpress.com/2022/08/04/pfam-website-decommission/).

Then, I suppose that the ProDy methods to retrieve data from Pfam will no longer be functional. To continue giving support for this useful feature the [IterPro API](https://github.com/ProteinsWebTeam/interpro7-api/tree/master/docs) can be used. 

In the case of adding a new module to ProDy for parsing InterPro, the old Pfam could make use of it and so maintaining the retro-compatibility for the current users.

jamesmkrieger:
Thanks for the information. We’ll see what we can do.

If you can help at all, that would be great

AmiteshBadkul:
Hi @jamesmkrieger, maybe the legacy version of the website (https://pfam-legacy.xfam.org/) could be used as a temporary fix (as the legacy version of the website will be taken down by Jan 2023).

jamesmkrieger:
Ok. Thanks

jamesmkrieger:
This temporary fix is now working and merged. You can now use pfam through the legacy website with the github master version of ProDy

> Hi @jamesmkrieger, maybe the legacy version of the website (https://pfam-legacy.xfam.org/) could be used as a temporary fix (as the legacy version of the website will be taken down by Jan 2023).

jamesmkrieger:
Update: we now have until the spring:
```
InterPro is the new home of Pfam
The Pfam website (pfam.xfam.org) was shut down on October 5th, but InterPro offers the same functionality and data.
A legacy version of Pfam is available at https://pfam-legacy.xfam.org/ but will not receive any updates and will be decommissioned in Spring 2023.
```
(https://www.ebi.ac.uk/interpro/)

jamesmkrieger:
The Interpro API doesn't seem to have quite as right a functionality as the pfam one, but maybe I haven't quite found everything. Nevertheless, I am making progress at porting over the functions to continue working well enough.

Best wishes
James

jamesmkrieger:
[...]

PR #1593, opened 2022-10-12
pfam to pfam-legacy

jamesmkrieger:
temporary fix for #1587


jamesmkrieger:
Now all the pfam things work again and there are tests for them too.

I've also added empty files ready for making tests for other database modules.

PR #1615, opened 2022-11-15
Pfam tests to not timeout

jamesmkrieger:
They could still hit a too many requests error but those are fewer


jamesmkrieger:
Actually, I'm removing them because they just cause too many problems.

PR #1617, opened 2022-11-15
Proper handling of fragments when bonds are absent

SHZ66:
**Old behavior:** calling `atom.getFragindex()` without bond info would lead to an error.
**New behavior:** calling `atom.getFragindex()` without bond info would return `None` and a warning.

The reason for this change is that raising an error for calling a `get` function seems to be too harsh. Besides, the new behavior provides an easier way (checking if `None` vs. catching an error) to programmatically check if the bond/fragment information is present.

### DIFF
--- a/prody/atomic/functions.py
+++ b/prody/atomic/functions.py
@@ -55,7 +55,7 @@ def saveAtoms(atoms, filename=None, **kwargs):
 
     if filename is None:
         filename = ag.getTitle().replace(' ', '_')
-    if filename.endswith('.npz'):
+    if filename.lower().endswith('.npz'):
         filename += '.ag.npz'
 
     attr_dict = {'title': title}
--- a/prody/atomic/functions.py
+++ b/prody/atomic/functions.py
@@ -55,7 +55,7 @@ def saveAtoms(atoms, filename=None, **kwargs):
 
     if filename is None:
         filename = ag.getTitle().replace(' ', '_')
-    if '.ag.npz' not in filename:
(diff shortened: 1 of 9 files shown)


prody/atomic/functions.py
```diff
@@ -55,7 +55,7 @@ def saveAtoms(atoms, filename=None, **kwargs):
 
     if filename is None:
         filename = ag.getTitle().replace(' ', '_')
-    if '.ag.npz' not in filename:
+    if filename.endswith('.npz'):
```
jamesmkrieger, line 58: this should still be `'ag.npz'`

jamesmkrieger (changes requested):
looks good except for one thing in atomic/functions

prody/atomic/functions.py
SHZ66: I think we should just check for ".npz" otherwise if the original name is "xxx.npz", it would rename it to "xxx.npz.ag.npz". The current change is consistent with the behavior in `saveModel` where it only checks for ".npz". Ultimately, as long as it is an NPZ file, the secondary suffix shouldn't matter.

However, I just realized that in many places we didn't consider the case sensitivity. So I am making changes in all those places.

jamesmkrieger (approved):

SHZ66:
@jamesmkrieger Thanks a lot for reviewing and approving the PR! Since all the tests are passed, I am merging this one as I've got another one coming :).

jamesmkrieger:
You're welcome

PR #1618, opened 2022-11-16
Error during writeout PQR file

changephilip:
```
File ~/envs/prody/lib/python3.9/site-packages/prody/proteins/pdbfile.py:1620, in writePQRStream(stream, atoms, **kwargs)
   1617         for item2 in sorted_sheet[i-num_strands:i]:
   1618             item2.append(num_strands)
-> 1620 num_strands = item[1]
   1621 for item2 in sorted_sheet[i-num_strands+1:]:
   1622     item2.append(num_strands)

NameError: name 'item' is not defined
```

**I am not sure but it works after modification**


jamesmkrieger:
It does seem to, yes.

It could have been that it was meant to be inside the loop though.

PR #1620, opened 2022-11-16
no sorting of strands in writing

jamesmkrieger:
Fixes both PDB and PQR writing where sheets have names inconsistent with sorting e.g., S1A, S2A, P1A, P2A, S1B in 2tbv

writePDBStream had an unintentional sorting from numpy unique that is now reversed by using sorted indices

writePQRStream had an actual sorting (that may have been to reproduce the writePDBStream behaviour) that is now removed too


jamesmkrieger:
We should probably have a test for writing SHEET lines to match ones in actual PDB files. At the moment, this would only work for the first parts of the lines but that's already something.

PR #1623, opened 2022-11-18
Allowing setFlags to take scalar values

SHZ66:
This makes `setFlags` behave the same way as `setData` and allows the following code to work:
```
import prody as pd

prot = pd.parsePDB('3tt1')

nodes = prot.ca
prot.setFlags("nmanodes", False)
nodes.setFlags("nmanodes", True)
sel = prot.select("nmanodes")
```

Otherwise, the user has to allocate the array with the correct length themselves without being able to use ProDy's own atom selection mechanism to set the flags.


SHZ66:
The checks seem to fail for unrelated reasons. I will run these checks again later.

jamesmkrieger:
> The checks seem to fail for unrelated reasons. I will run these checks again later.

Maybe support has been dropped for these versions

jamesmkrieger (approved):
Looks good to me. Thanks for fixing so many things. I guess I should make another release with all of them

SHZ66:
> > The checks seem to fail for unrelated reasons. I will run these checks again later.
> 
> Maybe support has been dropped for these versions

Yeah, it looks like the support might be dropped for those two ancient versions. Somehow `pyproject.toml` is missing and out of curiosity, I added it back to see if it would help with the issue with the checks (it did not).

Any reason `pyproject.toml` was removed in the first place? ProDy has `numpy` dependent C extensions so it has to import `numpy` in `setup.py`, which would require a toml file to make a pre-installation environment (without one, it would run into a [chicken-and-egg situation](https://stackoverflow.com/questions/62983756/what-is-pyproject-toml-file-for)).

jamesmkrieger:
I removed pyproject.toml because it was causing problems with biopython versions, especially for puthon 3.11

We never had one until recently so I thought it would be ok. Sorry for not checking with you

jamesmkrieger:
I guess the list of supported versions is out of date. I tried in #1624 and had the same problem

SHZ66:
> I removed pyproject.toml because it was causing problems with biopython versions, especially for puthon 3.11
> 
> We never had one until recently so I thought it would be ok. Sorry for not checking with you

Hmm, what kind of problem with biopython? Is there an issue associated with it? Maybe I can take a look.

Yeah, before we added `pyproject.toml` (or rather, before pip/python added this mechanism), we would ask the user to prepare an environment with numpy and biopython before installing ProDy. This kind of works except that it beats the purpose of having numpy listed as a dependency in `setup.py`. With `pyproject.toml`, whatever listed in it will be pre-installed for `setup.py` to work.

SHZ66:
> I guess the list of supported versions is out of date. I tried in #1624 and had the same problem

I am gonna merge this one and then try to fix the tests later in another PR.

jamesmkrieger:
> > I removed pyproject.toml because it was causing problems with biopython versions, especially for puthon 3.11
> > We never had one until recently so I thought it would be ok. Sorry for not checking with you
> 
> Hmm, what kind of problem with biopython? Is there an issue associated with it? Maybe I can take a look.

I think the version we’re asking for doesn’t exist compatible with Python 3.11 - do we still need this version?

jamesmkrieger:
@SHZ66, I've fixed the tests now in #1635. 

I decided Python 3.11 is too hard to support at the moment anyway. It causes problems with installing Biopython because of the install mechanism rules (see https://github.com/prody/ProDy/actions/runs/3941718837/jobs/6744482809).

PR #1633, opened 2023-01-06
Improved and reorganized module of Interactions

karolamik13:
New, improved version of the module for detecting interactions (hydrogen bonds, salt bridges, hydrophobic, repulsive ionic bonding, pi-stacking, and cation-pi) in proteins or between proteins and ligands. Now includes also analyzing of trajectories (dcd files).

### DIFF
--- a/prody/proteins/pdbfile.py
+++ b/prody/proteins/pdbfile.py
@@ -1057,13 +1057,13 @@ def _evalAltlocs(atomgroup, altloc, chainids, resnums, resnames, atomnames):
 #             '                               %5d\n')
 
 HELIXLINE = ('HELIX  {serNum:3d} {helixID:>3s} '
-             '{initResName:<3s}{initChainID:2s} {initSeqNum:4d}{initICode:1s} '
-             '{endResName:<3s}{endChainID:2s} {endSeqNum:4d}{endICode:1s}'
-             '{helixClass:2d}                               {length:5d}\n')
+             '{initResName:<3s}{initChainID:>2s} {initSeqNum:4d}{initICode:1s} '
+             '{endResName:<3s}{endChainID:>2s} {endSeqNum:4d}{endICode:1s}'
+             '{helixClass:2d}                               {length:5d}    \n')
 
 SHEETLINE = ('SHEET  {strand:3d} {sheetID:>3s}{numStrands:2d} '
-             '{initResName:3s}{initChainID:2s}{initSeqNum:4d}{initICode:1s} '
-             '{endResName:3s}{endChainID:2s}{endSeqNum:4d}{endICode:1s}{sense:2d} \n')
+             '{initResName:3s}{initChainID:>2s}{initSeqNum:4d}{initICode:1s} '
+             '{endResName:3s}{endChainID:>2s}{endSeqNum:4d}{endICode:1s}{sense:2d} \n')
 
 PDBLINE_LT100K = ('%-6s%5d %-4s%1s%-3s%2s%4d%1s   '
                   '%8.3f%8.3f%8.3f%6.2f%6.2f      '
--- a/prody/tests/proteins/test_pdbfile.py
+++ b/prody/tests/proteins/test_pdbfile.py
@@ -209,6 +209,8 @@ class TestWritePDB(unittest.TestCase):
         self.ag = parsePDB(self.pdb['path'])
         self.tmp = os.path.join(TEMPDIR, 'test.pdb')
 
+        self.ubi = parsePDB(DATA_FILES['1ubi']['path'], secondary=True)
+
         self.hex = parsePDB(DATA_FILES['hex']['path'])
         self.h36 = parsePDB(DATA_FILES['h36']['path'])
 
@@ -279,11 +281,20 @@ class TestWritePDB(unittest.TestCase):
 
         serial_99999_line = lines[99999]
         self.assertEqual(serial_99999_line[6:11], '99999',
-            'writePDB failed to write correct pre-h36 serial')        
+            'writePDB failed to write correct pre-h36 serial')
 
         serial_A0000_line = lines[100000]
         self.assertEqual(serial_A0000_line[6:11], 'A0000',
-            'writePDB failed to write correct h36 serial') 
+            'writePDB failed to write correct h36 serial')
+        
+    @dec.slow
+    @unittest.skipUnless(os.access(TEMPDIR, os.W_OK), msg)
+    def testWritingSecstrs(self):
+        """Test if output from writing secstrs is as expected."""
+
+        out = writePDB(self.tmp, self.ubi)
+        ubi_new = parsePDB(out, secondary=True)
+        self.assertListEqual(list(self.ubi.getSecstrs()), list(ubi_new.getSecstrs()))
 
     @dec.slow
     @unittest.skipUnless(os.access(TEMPDIR, os.W_OK), msg)
@@ -298,13 +309,6 @@ class TestWritePDB(unittest.TestCase):
                 'failed to write correct number of models')
             assert_equal(out.getCoords(), self.ag.getCoordsets(i),
                  'failed to write model {0} coordinates correctly'.format(i+1))
-
-    @dec.slow
-    def tearDown(self):
-        """Remove test file."""
-
-        if os.path.isfile(self.tmp):
-            os.remove(self.tmp)
 
     @dec.slow
     @unittest.skipUnless(os.access(TEMPDIR, os.W_OK), msg)
@@ -365,6 +369,13 @@ class TestWritePDB(unittest.TestCase):
         serial_A0000_line = lines[100000]
         self.assertEqual(serial_A0000_line[6:11], 'A0000',
             'writePDB failed to write correct h36 serial') 
+
+    @dec.slow
+    def tearDown(self):
+        """Remove test file."""
+
+        if os.path.isfile(self.tmp):
+            os.remove(self.tmp)
 
 
 class TestParsePDBHeaderAndAllModels(unittest.TestCase):


prody/proteins/interactions.py
```diff
@@ -0,0 +1,2030 @@
...
+    
+    :arg pdb: PDB file name
+    :type pdb: str
+
+    :arg method: Name of program which will be use to fix protein structure.
+            Two alternative options are available: 'openbabel' and 'pdbfixer'.
+            For either option additional software need to be installed:
+            'openbabel': OpenBabel
+            'pdbfixer': PDBFixer and OpenMM
+    :type method: str, by default is 'openbabel'
```
jamesmkrieger, line 115: I don't think this will work. `:type:` field is supposed to be for type and nothing else. Please move default up to a new line before it

prody/proteins/interactions.py
```diff
@@ -0,0 +1,2030 @@
...
+    
+    
+def calcHydrogenBonds(atoms, distA=3.0, angle=40, cutoff_dist=20, **kwargs):
+    """Compute hydrogen bonds for proteins and other molecules.
+    
+    :arg atoms: an Atomic object from which residues are selected
+    :type atoms: :class:`.Atomic`
+    
+    :arg distA: non-zero value, maximal distance between donor and acceptor.
+    :type distA: int or float, default is 3.0
```
jamesmkrieger, line 176: same here

prody/proteins/interactions.py
```diff
@@ -0,0 +1,2032 @@
...
+    
+    :arg angle: non-zero value, maximal (180 - D-H-A angle) (donor, hydrogen, acceptor).
+    :type distA: int or float, default is 40.
+    
+    :arg cutoff_dist: non-zero value, interactions will be found between atoms with index differences
```
jamesmkrieger, line 181: I fixed this here. Please apply to calcChHydrogenBonds or better still remove it from there and say that kwargs are the same ones

prody/proteins/interactions.py
```diff
@@ -0,0 +1,2034 @@
...
+    
+    contacts = findNeighbors(atoms, distA)
+    short_contacts = cleanNumbers(contacts)
+    pairList = [] # list with Donor-Hydrogen-Acceptor(indices)-distance-Angle
+    
+    LOGGER.info('Calculating hydrogen bonds.')
+    for nr_i,i in enumerate(short_contacts):
+        # Removing those close contacts which are between neighbour atoms
+        if i[1] - cutoff_dist < i[0] < i[1] + cutoff_dist:
+            pass
```
jamesmkrieger, line 217: you can use `continue` to continue on to the next round of the loop and then you don't need an `else` in the next line

prody/proteins/interactions.py
```diff
@@ -0,0 +1,2034 @@
...
+            raise TypeError('coords must be an object '
+                            'with `getCoords` method')
+    
+    donors = kwargs.get('donors', ['N', 'O', 'S', 'F'])
+    acceptors = kwargs.get('acceptors', ['N', 'O', 'S', 'F'])
+    
+    if atoms.hydrogen == None or atoms.hydrogen.numAtoms() < 10:
+        LOGGER.info("Provide structure with hydrogens or install Openbabel to add missing hydrogens using addHydrogens(pdb_name) first.")
+    
+    contacts = findNeighbors(atoms, distA)
```
jamesmkrieger, line 209: it probably makes sense to use atoms.heavy here because you're analysing hydrogens separately below. It's quite inefficient to loop through so many contacts including hydrogens otherwise

prody/proteins/interactions.py
```diff
@@ -0,0 +1,2034 @@
...
+
+
+class Interactions(object):
+
+    """Class for Interaction analysis of proteins."""
+
+    def __init__(self, title='Unknown'):
+        self._title = str(title).strip()
+           
+    def setTitle(self, title):
```
jamesmkrieger, line 1406: why does setTitle initialise all these things? It's not even used by any other function. It would be much better to do this in __init__ or calcProteinInteractions. Or perhaps, you could call setTitle from calcProteinInteractions

prody/proteins/interactions.py
```diff
@@ -0,0 +1,2043 @@
...
+
+        LOGGER.info('Legend: hb-hydrogen bond, sb-salt bridge, rb-repulsive ionic bond, ps-Pi stacking interaction,'
+                                                    ' pc-Cation-Pi interaction, hp-hydrophobic interaction')    
+        from toolz.curried import count
+        LOGGER.info('The biggest number of interactions: {}'.format(max(map(count, ListOfInteractions))))
+        
+        return ListOfInteractions_list
+        
+
+    def showFrequenctInteractions(self, cutoff=5, **kwargs):
```
jamesmkrieger, line 1723: this function name has a typo. There shouldn't be a c before the t

prody/proteins/interactions.py
```diff
@@ -0,0 +1,2043 @@
...
+        (4) Pi stacking interactions
+        (5) Pi-cation interactions
+        (6) Hydrophobic interactions
+
+For protein-ligand interactions (3) is replaced by water bridges.
+"""
+
+__author__ = 'Karolina Mikulska-Ruminska'
+__credits__ = ['James Krieger','Karolina Mikulska-Ruminska']
+__email__ = ['karolami@pitt.edu', 'KRIEGERJ@pitt.edu']
```
jamesmkrieger, line 18: it's probably better not to use our pitt emails when we probably won't have them that much longer. We could use my gmail maybe

prody/proteins/interactions.py
jamesmkrieger: types should also be separated by commas not `or`

prody/proteins/interactions.py
```diff
@@ -0,0 +1,2043 @@
...
+    """Interactions with metal ions (includes water, ligands and other ions).
+        
+    :arg atoms: an Atomic object from which residues are selected
+    :type atoms: :class:`.Atomic`
+    
+    :arg distA: non-zero value, maximal distance between ion and residue.
+    :type distA: int or float, default is 3.0.
+    
+    :arg extraIons: ions to be included in the analysis.
+    :type extraIons: list of str
```
jamesmkrieger, line 706: list of str should be commented in the main description and only list should be stated as the type

prody/proteins/interactions.py
```diff
@@ -0,0 +1,2043 @@
...
+    
+    :arg excluded_ions: ions which should be excluded from the analysis.
+    :type excluded_ions: list of str """
```
jamesmkrieger, line 709: same here

prody/proteins/interactions.py
```diff
@@ -0,0 +1,2043 @@
...
+    for the same system. Two conformations can be compared.
+    
+    :arg data1: list with interactions from calcHydrogenBonds() or other types
+    :type data1: list
+ 
+    :arg data2: list with interactions from calcHydrogenBonds() or other types
+    :type data2: list 
+    
+    Example of usage: 
+    >>> atoms1 = parsePDB('PDBfile1.pdb').select('protein')
```
jamesmkrieger, line 987: sometimes there are triple > like here and sometimes double ones. VS code highlights triple ones in purple so maybe they should all be like that. Perhaps check rst guidelines and/or other modules

prody/proteins/interactions.py
```diff
@@ -0,0 +1,2043 @@
...
+    try:
+        output = kwargs['output']
+    except:
+        output = atoms.getTitle()+'_interaction.tcl'
+    
+    tcl_file = open(output, 'w') 
+    
+    
+    def TCLforSingleInteaction(interaction, color='blue', tcl_file=tcl_file):
+        """Protein interactions from ProDy for a single Interaction"""
```
jamesmkrieger, line 1287: I know this is an internal function, but it could still be good to have better documentation of what this means

prody/proteins/interactions.py
```diff
@@ -0,0 +1,2043 @@
...
+            if 'output' in kwargs:
+                writePDB(kwargs['output'], atoms, **kw)  
+                LOGGER.info('PDB file saved.')
+            else:
+                writePDB('output', atoms, **kw)
+                LOGGER.info('PDB file saved.')
+        except: LOGGER.info('There is a problem.')
+        
+
+    def getFrequentInteractions(self, contacts_min=3):
```
jamesmkrieger, line 1678: perhaps this could say frequent interactors instead

prody/proteins/interactions.py
```diff
@@ -0,0 +1,2043 @@
...
+            (3) RepulsiveIonicBonding 
+            (4) Pi stacking interactions
+            (5) Pi-cation interactions
+            (6) Hydrophobic interactions
+        
+        :arg atoms: an Atomic object from which residues are selected
+        :type atoms: :class:`.Atomic`
+        
+        :arg trajectory: DCD filename
+        :type trajectory: str
```
jamesmkrieger, line 1811: it could be good to set it like this. At the moment, we are using a Trajectory object and having the problem with getting to the end of the frames

prody/proteins/interactions.py
```diff
@@ -0,0 +1,2043 @@
...
+            coords = (atoms._getCoords() if hasattr(atoms, '_getCoords') else
+                        atoms.getCoords())
+        except AttributeError:
+            try:
+                checkCoords(coords)
+            except TypeError:
+                raise TypeError('coords must be an object '
+                                'with `getCoords` method')
+        
+        if not isinstance(trajectory, (TrajBase, Ensemble, Atomic)):
```
jamesmkrieger, line 1827: this doesn't work. If you enumerate an Atomic object then you get atoms not frames. You probably need to convert it to an ensemble

jamesmkrieger: something like the following:
```
        if isinstance(trajectory, Atomic):
            trajectory = Ensemble(trajectory)
```

jamesmkrieger: I added this here now but it still needs adding to other places

prody/proteins/interactions.py
```diff
@@ -0,0 +1,2047 @@
...
+        coords = (atoms._getCoords() if hasattr(atoms, '_getCoords') else
+                    atoms.getCoords())
+    except AttributeError:
+        try:
+            checkCoords(coords)
+        except TypeError:
+            raise TypeError('coords must be an object '
+                            'with `getCoords` method')
+        
+    if not isinstance(trajectory, (TrajBase, Ensemble, Atomic)):
```
jamesmkrieger, line 790: please convert atomic before enumerating

prody/proteins/interactions.py
```diff
@@ -0,0 +1,2047 @@
...
```
jamesmkrieger, line 824: please convert atomic before enumerating

prody/proteins/interactions.py
```diff
@@ -0,0 +1,2047 @@
...
```
jamesmkrieger, line 892: please convert atomic before enumerating

prody/proteins/interactions.py
```diff
@@ -0,0 +1,2047 @@
...
```
jamesmkrieger, line 926: please convert atomic before enumerating

prody/proteins/interactions.py
```diff
@@ -0,0 +1,2047 @@
...
```
jamesmkrieger, line 960: please convert atomic before enumerating

prody/proteins/interactions.py
karolamik13: Done

prody/proteins/interactions.py
karolamik13: Done

prody/proteins/interactions.py
karolamik13: Changed

prody/proteins/interactions.py
karolamik13: Done

prody/proteins/interactions.py
karolamik13: Moved to init

prody/proteins/interactions.py
karolamik13: Done

prody/proteins/interactions.py
```diff
@@ -761,31 +761,29 @@ def calcDisulfideBonds(atoms, distA=2.5):
                             'with `getCoords` method')
 
     try:
-        atoms_SG = atoms.select('protein and resname CYS')
-    except AttributeError:
-        try:
-            checkCoords(atoms_SG)
-        except TypeError:
-            raise TypeError('Lack of cysteines in the structure.')
+        atoms_SG = atoms.select('protein and resname CYS and name SG')
```
jamesmkrieger, line 764: You could make the selection outside a try/except and then use an if to check if there are cys sg atoms.

prody/proteins/interactions.py
```diff
@@ -761,31 +761,29 @@ def calcDisulfideBonds(atoms, distA=2.5):
...
-    if len(DisulfideBonds_list2) != 0:
-        return DisulfideBonds_list2
-    else:
-        LOGGER.info('Lack of disulfide bonds in the structure.')
+        if len(DisulfideBonds_list2) != 0:
+            return DisulfideBonds_list2
+        else:
+            LOGGER.info('Lack of disulfide bonds in the structure.')
     
+    except:
```
jamesmkrieger, line 782: it's better to use a more specific except but, as I said above, I don't think you need one at all

prody/proteins/interactions.py
```diff
@@ -867,14 +865,31 @@ def calcProteinInteractions(atoms, **kwargs):
...
     :arg atoms: an Atomic object from which residues are selected
     :type atoms: :class:`.Atomic`
         
-    :arg trajectory: DCD filename
-    :type trajectory: str"""
+    :arg trajectory: trajectory file
+    :type trajectory: class:`.Trajectory`
+
+    :arg distA: non-zero value, maximal distance between donor and acceptor.
+    :type distA: int, float, default is 3.0
```
jamesmkrieger, line 878: have you checked that putting the default like this works yet?

prody/proteins/interactions.py
```diff
@@ -867,14 +865,31 @@ def calcProteinInteractions(atoms, **kwargs):
...
+    
+    :arg cutoff_dist: non-zero value, interactions will be found between atoms with index differences
+        that are higher than cutoff_dist.
+        default is 20 atoms.
+    :type cutoff_dist: int
+    
+    :arg selection: selection string
+    :type selection: str
+    
+    :arg selection2: selection string
```
jamesmkrieger, line 891: it would probably be good to explain what this is better

prody/proteins/interactions.py
```diff
@@ -867,14 +865,31 @@ def calcProteinInteractions(atoms, **kwargs):
     return AllInteractions
 
 
-def calcHydrogenBondsDCD(atoms, trajectory, distA=3.0, angle=40, cutoff_dist=20, **kwargs):   
+def calcHydrogenBondsDCD(atoms, trajectory=None, distA=3.0, angle=40, cutoff_dist=20, **kwargs):
```
jamesmkrieger, line 868: you should probably rename the function if it doesn't need a trajectory. You could even have all the functions as general ones that can take a trajectory or not.

prody/proteins/interactions.py
```diff
@@ -885,29 +900,53 @@ def calcHydrogenBondsDCD(atoms, trajectory, distA=3.0, angle=40, cutoff_dist=20,
...
+    else:
+        if atoms.numCoordsets() > 1:
+            for i in range(len(atoms.getCoordsets())):
+                LOGGER.info('Model: {0}'.format(i))
+                atoms.setACSIndex(i) 
+                protein = atoms.select('protein')
+                hydrogen_bonds = calcHydrogenBonds(protein, distA, angle, cutoff_dist, **kwargs)
+                HBs_all.append(hydrogen_bonds)
+        else:
+            LOGGER.info('Include trajectory or use multiple PDB file.')
```
jamesmkrieger, line 927: multi-state pdb file

prody/proteins/interactions.py
```diff
@@ -885,29 +900,53 @@ def calcHydrogenBondsDCD(atoms, trajectory, distA=3.0, angle=40, cutoff_dist=20,
...
+            
     return HBs_all
 
 
-def calcSaltBridgesDCD(atoms, trajectory, distA=4.5):
+def calcSaltBridgesDCD(atoms, trajectory=None, distA=4.5, **kwargs):
```
jamesmkrieger, line 932: again, not dcd but something else and explain. I'd still suggest supporting an Ensemble too

prody/proteins/interactions.py
```diff
@@ -918,29 +957,52 @@ def calcSaltBridgesDCD(atoms, trajectory, distA=4.5):
...
+    SBs_all = []
 
-    if isinstance(trajectory, Atomic):
-        trajectory = Ensemble(trajectory)        
+    if trajectory is not None:
+        if isinstance(trajectory, Atomic):
+            trajectory = Ensemble(trajectory)        
                         
-    SBs_all = []
+        trajectory.reset()
```
jamesmkrieger, line 966: maybe store the current active frame and set it back to there at the end

prody/proteins/interactions.py
```diff
@@ -885,29 +900,53 @@ def calcHydrogenBondsDCD(atoms, trajectory, distA=3.0, angle=40, cutoff_dist=20,
...
         
-    for j0, frame0 in enumerate(trajectory):  
-        LOGGER.info('Frame: {0}'.format(j0))
-        protein = atoms.select('protein')
-        hydrogen_bonds = calcHydrogenBonds(protein, distA, angle, cutoff_dist, **kwargs)
-        HBs_all.append(hydrogen_bonds)
-        
+        for j0, frame0 in enumerate(trajectory):  
+            LOGGER.info('Frame: {0}'.format(j0))
+            protein = atoms.select('protein')
```
jamesmkrieger, line 914: This doesn't feel like it would work, but maybe I'm wrong. I don't see anywhere where you set the coordinates of the frame to be related to the atoms.

prody/proteins/interactions.py
jamesmkrieger: probably you do support Ensemble but you need to have docs and a type check

prody/proteins/interactions.py
```diff
@@ -918,29 +957,52 @@ def calcSaltBridgesDCD(atoms, trajectory, distA=4.5):
         except TypeError:
             raise TypeError('coords must be an object '
                             'with `getCoords` method')
+    SBs_all = []
 
-    if isinstance(trajectory, Atomic):
-        trajectory = Ensemble(trajectory)        
+    if trajectory is not None:
+        if isinstance(trajectory, Atomic):
```
jamesmkrieger, line 963: you could have an elif here that if it's not Trajectory or Ensemble then raise a type error

jamesmkrieger (changes requested):
generally seems good but still some suggestions

prody/proteins/interactions.py
```diff
@@ -1921,10 +1921,12 @@ def buildInteractionMatrix(self, **kwargs):
         return InteractionsMap
 
 
-    def showInteractions(self, **kwargs):
+    def showInteractions(self, filename=None, **kwargs):
```
jamesmkrieger, line 1924: I don't think this is the prody way. People can save figures themselves and do more complicated things like subplots too.

prody/proteins/interactions.py
```diff
@@ -2212,57 +2222,151 @@ def calcProteinInteractionsDCD(self, atoms, trajectory, filename=None, **kwargs)
...
+            return self._interactions_dcd
+        else:
+            sele_inter = []
+            for i in self._interactions_dcd:
+                for nr_j,j in enumerate(i):
+                    sele_inter.append(selectionByKwargs(i[nr_j], self._atoms, **kwargs))
+            self._interactions_dcd = sele_inter
+            return self._interactions_dcd
+        
+        if 'filename' != None:
```
jamesmkrieger, line 2247: this should go first. Otherwise, this code is never run because both the parts above have a return

prody/proteins/interactions.py
```diff
@@ -2212,57 +2222,151 @@ def calcProteinInteractionsDCD(self, atoms, trajectory, filename=None, **kwargs)
...
-    def getHydrophobic(self):
-        """Return the list of hydrophobic interactions computed from DCD trajectory"""
-        
-        return self._hps_dcd
+        if len(kwargs) == 0:
+            return self._hbs_dcd
+        else:
+            sele_inter = []
+            for nr_i,i in enumerate(self._hbs_dcd):
+                sele_inter.append(selectionByKwargs(self._hbs_dcd[nr_i], self._atoms, **kwargs))
```
jamesmkrieger, line 2277: this function should have a better name that says something about interactions

prody/proteins/interactions.py
```diff
@@ -2334,8 +2438,11 @@ def parseInteractions(self, filename):
         
         return data
     
-    def getTimeInteractions(self, **kwargs):    
-        """Return a bar plots with the number of interactions per each frame """
+    def getTimeInteractions(self, filename=None, **kwargs):    
+        """Return a bar plots with the number of interactions per each frame.
+        
+        :arg filename: PNG file name
```
jamesmkrieger, line 2444: again, I don't think we need this

jamesmkrieger (approved):
if you want this then go ahead, but it's not the prody way

prody/proteins/interactions.py
```diff
@@ -2222,57 +2222,151 @@ def calcProteinInteractionsDCD(self, atoms, trajectory, filename=None, **kwargs)
...
+                for nr_j,j in enumerate(i):
+                    sele_inter.append(selectionByKwargs(i[nr_j], self._atoms, **kwargs))
+            self._interactions_dcd = sele_inter
+            return self._interactions_dcd
+        
+        if 'filename' != None:
             with open(kwargs['filename']+'.dat', 'wb') as f:
                 pickle.dump(self._interactions_dcd, f)  
             LOGGER.info('File with interactions saved.')
         else: pass
```
jamesmkrieger, line 2251: this is not necessary

jamesmkrieger (changes requested):
There's a significant problem here that needs to be fixed. The filename part wouldn't get run because the if/else block has returns in both parts. They are both the same return like so it can be moved down to the end

prody/proteins/interactions.py
jamesmkrieger: I've done this now

prody/proteins/interactions.py
jamesmkrieger: It might be that we can't check that without building the website, but maybe there's another way

prody/proteins/interactions.py
jamesmkrieger: I've done this now

prody/proteins/interactions.py
jamesmkrieger: I've fixed this

jamesmkrieger:
> There's a significant problem here that needs to be fixed. The filename part wouldn't get run because the if/else block has returns in both parts. They are both the same return like so it can be moved down to the end

I've fixed this now

The other thing I'm concerned about is selecting protein from atoms that seem to know about frames

prody/proteins/interactions.py
karolamik13: I did that because in jupyter notebook I was not able to save it. It is probably because of plt.show() which is included in the method.

prody/proteins/interactions.py
karolamik13: I tested each change that I made in the code. But I will double-check.

prody/proteins/interactions.py
jamesmkrieger: There shouldn't be a plt.show() either. ProDy has a way for checking what figures are open and so on, separately through auto_show. See showEllipsoid or showFractVars for examples

prody/proteins/interactions.py
jamesmkrieger: ok

prody/proteins/interactions.py
karolamik13: Yes, I do support Ensemble. If you use PDB with multiple frames as atoms and do not include trajectory. It will compute the results for multiple frames.

prody/proteins/interactions.py
karolamik13: 'trajectory' is for dcd files only. If you use pdb file with multiple pdb frames, then the function will compute it all as Ensemble.

prody/proteins/interactions.py
karolamik13: I will do that. I was planning to do it in the tutorial, but I can do it here as well. 
selection is for any selection you want, such as chain A or chain B and resid 50 to 100, etc.
If you want to have interactions ONLY between two chains you can use selection and selection2:
selection='chain A', selection2='chain B'

prody/proteins/interactions.py
jamesmkrieger: The way to do this would be to have the following line before the reset
```
nfi = trajectory._nfi # next frame index
```
and the following one at the end of the function before the return
```
trajectory._nfi = nfi
```

jamesmkrieger: I'd probably also provide a kwarg for whether the user wants to reset back to the beginning. They may want to analyse from frame 3 onwards for example to skip some equilibration period.

prody/proteins/interactions.py
jamesmkrieger: No, I meant actually providing an Ensemble as input. I haven't tested it but I'm pretty sure it should work if you can convert the atoms to one inside

prody/proteins/interactions.py
karolamik13: I modified this function a few times with different try/except etc. I think I also did the way you described, but after tests on different structures, I had to modify it. For some reason, it was working badly on one structure that I tested. The code which we have now was working correctly on all structures it was tested.

prody/proteins/interactions.py
jamesmkrieger: but, trajectory can be None and it can use multi-frame pdb files. 

(It actually can't use multiple pdb files without *atoms and buildPDBEnsemble, which I'm sure you don't want)

You could call it calcHydrogenBondsTrajectory because it can use multi-frame Atomic not just dcd Trajectory.

prody/proteins/interactions.py
jamesmkrieger: You could have something like the following
```
        if isinstance(trajectory, Atomic):
            trajectory = Ensemble(trajectory)
        elif not isinstance(Trajectory, (Trajectory, Ensemble):
            raise TypeError('Trajectory should be an Atomic, Trajectory or Ensemble object, or None')
```

prody/proteins/interactions.py
jamesmkrieger: Not all users will read the tutorial and those that do may forget. It's still good to say that selection2 is optional and that a sphere around selection will be provided if it isn't provided.

prody/proteins/interactions.py
jamesmkrieger: ok

jamesmkrieger (approved):

prody/proteins/interactions.py
```diff
@@ -0,0 +1,2478 @@
...
+    else:
+        final = list_of_interactions
+    return final
+
+
+def addHydrogens(pdb, method='openbabel', pH=7.0):    
+    """Function will add hydrogens to the protein and ligand structure using Openbabel [NO11]_
+    or PDBFixer with OpenMM.
+    
+    :arg pdb: PDB file name
```
jamesmkrieger, line 126: we usually use pdb when there's an option to use a pdb id and parse it. Otherwise, it's probably better to call it filename

prody/proteins/interactions.py
```diff
@@ -0,0 +1,2478 @@
...
+    if method == 'openbabel':
+        try:
+            #import openbabel
+            from openbabel import openbabel
+            obconversion = openbabel.OBConversion()
+            obconversion.SetInFormat("pdb")
+            mol = openbabel.OBMol()
+            obconversion.ReadFile(mol, pdb)
+            mol.AddHydrogens()
+            obconversion.WriteFile(mol, 'addH_'+pdb)
```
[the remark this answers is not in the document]
jamesmkrieger: openbabel also doesn't like cif files

prody/proteins/interactions.py
karolamik13: FrequentInteractions instead of getFrequentInteractions, or you mean contact_min to frequent_interactions?

jamesmkrieger: no. Like we discussed with Ivet, its the residues and not the interactions that are frequent and therefore we need a different name for the function. My suggestion was getFrequentInteractors with an ORS and not getFrequentInteractions with an IONS

karolamik13: Ok. Now I see. Thanks.

jamesmkrieger: You're welcome

prody/proteins/interactions.py
karolamik13: Done

prody/proteins/interactions.py
```diff
@@ -2219,36 +2422,38 @@ def calcProteinInteractionsDCD(self, atoms, trajectory, filename=None, **kwargs)
...
-        :type selection2: str """
+        :type selection2: str 
+            
+        Selection:
+        If we want to select interactions for the particular residue or group of residues: 
+            selection='chain A and resid 1 to 50'
+        If we want to study chain-chain interactions:
+            selection='chain A', selection2='chain B'  """
 
         if len(kwargs) != 0:
```
jamesmkrieger, line 2433: If filename is in the kwargs then this may not work as expected. Perhaps do the following first:

filename = kwargs.pop(“filename”, None)

prody/proteins/interactions.py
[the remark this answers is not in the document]
jamesmkrieger: > doesn't work to add the prefix to the whole filename:
> 
> ```
> ==============================
> *** Open Babel Error  in ReadMolecule
>   ERROR: not a valid PDB file
> 
> ==============================
> *** Open Babel Error  in WriteFile
>   Cannot write to addH_/home/jkrieger/ScipionUserData/projects/D614G_prody/Runs/000554_ProtImportSetOfAtomStructs/extra/7kec.cif
> ```
> 
> you need to use os.path.split as follows:
> 
> ```
> os.path.join(os.path.split(pdb)[0], "addH_" + os.path.split(pdb)[1])
> ```

I've fixed this now.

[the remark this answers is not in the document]
jamesmkrieger: > openbabel also doesn't like cif files

maybe you could use parsePDB and writePDB to give it a pdb file instead. Otherwise, you should probably add a note about that to the docs and raise a ValueError

prody/proteins/interactions.py
karolamik13: Done. Now it is filterInteractions()

prody/proteins/interactions.py
karolamik13: Done.

prody/proteins/interactions.py
karolamik13: I improved those which were not specific.

prody/proteins/interactions.py
karolamik13: It was working fine, but I changed it. Now all is in kwargs.

prody/proteins/interactions.py
karolamik13: It is working correctly.

prody/proteins/interactions.py
karolamik13: Done. Also, I added start_frame and stop_frame. Now everyone may choose which frame or model in multi-model PDB analyze.

jamesmkrieger: sounds good!

jamesmkrieger (commented):
looks very good. It could be nice though to have a separate function for handling the start and stop frames and data types so there isn't so much duplicated code to maintain. You essentially have the same code for analysing the interactions in each loop. 

This function could even have an argument that is the function name for the particular interactions to analyse and then the other functions would just be wrappers for it that provide the inputs and function name.

prody/proteins/interactions.py
```diff
@@ -718,8 +718,9 @@ def calcPiCation(atoms, **kwargs):
             sele1 = atoms.select('resid '+str(i[0])+' and chain '+i[1]+' and '+aromatic_dic[sele1_name[0]])
             sele2 = atoms.select('(same residue as exwithin '+str(distA)+' of center) and resname ARG LYS and noh and not backbone and not name NE "C.*"', 
                                center=calcCenter(sele1.getCoords()))
-        except:
-            LOGGER.info("Missing atoms from the side chains of the structure. Use PDBFixer.")
+        except ImportError:
```
jamesmkrieger, line 721: no, this is not an import error for the except because you didn't try to import anything. There could be an index error or a dictionary key error or an error from calcCenter not having anything in sele1

prody/proteins/interactions.py
```diff
@@ -1732,8 +1733,8 @@ def calcLigandInteractions(atoms, **kwargs):
                 pdb_name = pdb_name[:-4]+'_addH.pdb'
                 atoms = parsePDB(pdb_name)
                 LOGGER.info("Lack of hydrogens in the structure. Hydrogens have been added.")
-        except: 
-            LOGGER.info("Install Openbabel to add missing hydrogens or provide structure with hydrogens")
+        except ImportError:
```
jamesmkrieger, line 1840: this one could be an import error but there are probably other errors that could happen too. It would also be better for addHydrogens or whatever it is called now to handle the import errors itself.

prody/proteins/interactions.py
```diff
@@ -1774,8 +1775,8 @@ def calcLigandInteractions(atoms, **kwargs):
 
         return Ligands, analyzedLigand
 
-    except:
-        LOGGER.info("Install Openbabel and PLIP.")
+    except ImportError:
```
jamesmkrieger, line 1882: again there's quite a bit that could happen besides the import error. If it's the import error that you want to catch then you should handle it right after trying the import

jamesmkrieger (changes requested):
I don't think ImportError is right for any of these in the current way that they are written

prody/proteins/interactions.py
karolamik13: That was fine, yes? (below)
        except:
            LOGGER.info("Missing atoms from the side chains of the structure. Use PDBFixer.")
 I will change it here and in other places.

jamesmkrieger: Can you still make it an error not a LOGGER.info? Otherwise, yes, I guess it's ok. 

You could do
```
except:
    ValueError("Missing atoms from the side chains of the structure. Use addMissingAtoms.")
```

prody/proteins/interactions.py
karolamik13: Done.

prody/proteins/interactions.py
karolamik13: Ok. I will change that.

prody/proteins/interactions.py
karolamik13: Done.

jamesmkrieger (approved):

prody/proteins/interactions.py
```diff
@@ -1032,49 +1091,8 @@ def calcHydrogenBondsTrajectory(atoms, trajectory=None, **kwargs):
...
-        if atoms.numCoordsets() > 1:
-            for i in range(len(atoms.getCoordsets()[start_frame:stop_frame])):
-                LOGGER.info('Model: {0}'.format(i+start_frame))
-                atoms.setACSIndex(i+start_frame)
-                protein = atoms.select('protein')
-                hydrogen_bonds = calcHydrogenBonds(protein, **kwargs)
-                HBs_all.append(hydrogen_bonds)
-        else:
-            LOGGER.info('Include trajectory or use multi-model PDB file.')
+    HBs_all = calcInteractionsMultipleFrames(atoms, 'HBs', trajectory, **kwargs)
```
jamesmkrieger, line 1098: you could save on a variable and just say 
`return calcInteractionsMultipleFrames(atoms, 'HBs', trajectory, **kwargs)`

prody/proteins/interactions.py
```diff
@@ -2410,7 +2416,7 @@ def showFrequentInteractors(self, cutoff=5, **kwargs):
             fig = plt.figure(num=None, figsize=(12,6), facecolor='w')
         
         y_pos = np.arange(len(y))
-        show = plt.bar(y_pos, x, align='center', alpha=0.5, color='blue', **kwargs)
```
jamesmkrieger: why? Users may want to set bar kwargs such as width or shading

prody/proteins/interactions.py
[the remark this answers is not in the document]
jamesmkrieger: I fixed it by raising an error

prody/proteins/interactions.py
karolamik13: I know, but kwargs are also needed to add to the dictionary non-standard residues to include them.        
for key, value in kwargs.items():
            aa_dict[key] = value
(line 2393) 
I want to add kwargs in plt.bar as well, but they will be in conflict with non-standard names. Any idea how to solve it? I have for example HSD1, HSD2 etc, and my colleague CYX. Other people may also have non-standard residues/names.

jamesmkrieger: ok, fair point. Probably, it's too difficult to control

karolamik13: Non-standard may have 3 or 4 letters. Bigger will not fit the PDB standards. If kwargs for plt.bar will have more than >4 Ietters then maybe like this?

jamesmkrieger: Unfortunately not. They have ones like log and xy. Let's just leave it non-customisable, but maybe put something in the docs.

karolamik13: Ok.

prody/proteins/interactions.py
```diff
@@ -1490,14 +1490,20 @@ def showInteractionsGraph(statistics, **kwargs):
...
     
     
-def calcStatisticsInteractions(data):
-    """Return the statistics of interactions from DCD trajectory or multi-model PDB including
-    the number of counts for each residue pair, 
-    average distance of interactions for each pair [in Ang] and standard deviation.
+def calcStatisticsInteractions(data, **kwargs):
+    """Return the statistics of interactions from PDB Ensemble or trajectory including:
+    (1) the weight for each residue pair: corresponds to the number of counts divided by the 
+    number of frames (values >1 are obtained when residue pair creates multiple contacts;
```
jamesmkrieger, line 1496: This needs a closing bracket

prody/proteins/interactions.py
```diff
@@ -1490,14 +1490,20 @@ def showInteractionsGraph(statistics, **kwargs):
...
+def calcStatisticsInteractions(data, **kwargs):
+    """Return the statistics of interactions from PDB Ensemble or trajectory including:
+    (1) the weight for each residue pair: corresponds to the number of counts divided by the 
+    number of frames (values >1 are obtained when residue pair creates multiple contacts; 
+    (2) average distance of interactions for each pair [in Ang] and (3) standard deviation [Ang.].
         
     :arg data: list with interactions from calcHydrogenBondsTrajectory() or other types
     :type data: list
     
+    :arg weight_cuoff: value above which results will be displayed
```
jamesmkrieger, line 1502: Cutoff is missing the t

jamesmkrieger: Same below and in code

prody/proteins/interactions.py
```diff
@@ -1521,16 +1529,18 @@ def calcStatisticsInteractions(data):
             stats[element] = {
                 "stddev": np.round(np.std(values),2),
                 "mean": np.round(np.mean(values),2),
-                "count": len(values)
+                "weight": np.round(float(len(values))/len(data), 2)
```
jamesmkrieger, line 1532: It’s better not to round this for returning values to users and let them round it or not as they wish

docs/docs
```diff
@@ -1 +1 @@
-../docs
\ No newline at end of file
+ProDy/docs
```
jamesmkrieger, line 1: why is this changed?

prody/proteins/interactions.py
```diff
@@ -0,0 +1,3207 @@
...
+    
+    Hydrophobic_list = []  
+    atoms_hydrophobic = atoms.select('resname ALA VAL ILE MET LEU PHE TYR TRP')
+    hydrophobic_resids = list(set(zip(atoms_hydrophobic.getResnums(), atoms_hydrophobic.getChids())))
+
+    if atoms.aromatic is None:
+        return []
+    
+    aromatic_nr = list(set(zip(atoms.aromatic.getResnums(),atoms.aromatic.getChids())))   
+    aromatic = list(set(zip(atoms.aromatic.getResnames())))
```
jamesmkrieger, line 764: same here

jamesmkrieger:
Here is the test for speed-up:

Code (note starting structure and order):
```
from prody import *
import time

#atoms = parsePDB("7ms7", compressed=False)
#atoms2 = parsePDB(addMissingAtoms("7ms7.pdb", method='pdbfixer'))
atoms2 = parsePDB("addH_7ms7.pdb")

interactions = Interactions('')

start_time = time.time()
pistacks = calcPiStacking(atoms2.protein)
print("--- %s seconds ---" % (time.time() - start_time))

start_time = time.time()
hbs = calcHydrogenBonds(atoms2.protein)
print("--- %s seconds ---" % (time.time() - start_time))

start_time = time.time()
hph = calcHydrophobic(atoms2.protein)
print("--- %s seconds ---" % (time.time() - start_time))

start_time = time.time()
rbs = calcRepulsiveIonicBonding(atoms2.protein)
print("--- %s seconds ---" % (time.time() - start_time))

start_time = time.time()
dbs = calcDisulfideBonds(atoms2.protein)
print("--- %s seconds ---" % (time.time() - start_time))

start_time = time.time()
sbs = calcSaltBridges(atoms2.protein)
print("--- %s seconds ---" % (time.time() - start_time))

start_time = time.time()
pcs = calcPiCation(atoms2.protein)
print("--- %s seconds ---" % (time.time() - start_time))

start_time = time.time()
mbs = calcMetalInteractions(atoms2, extraIons=['ZN'])
print("--- %s seconds ---" % (time.time() - start_time))
```

Original:
```
--- 1.0006673336029053 seconds ---
--- 1.1017322540283203 seconds ---
--- 5.97992205619812 seconds ---
--- 0.043082237243652344 seconds ---
--- 0.05671882629394531 seconds ---
--- 0.09984111785888672 seconds ---
--- 0.6527218818664551 seconds ---
--- 0.004638195037841797 seconds ---
```

Improved:
```
--- 0.4951136112213135 seconds ---
--- 0.8739936351776123 seconds ---
--- 5.865622520446777 seconds ---
--- 0.04063129425048828 seconds ---
--- 0.05795001983642578 seconds ---
--- 0.10210824012756348 seconds ---
--- 0.6715052127838135 seconds ---
--- 0.005453586578369141 seconds ---
```

Main speed-ups are in calcPiStacking (~2x)

prody/proteins/interactions.py
```diff
@@ -1478,18 +1489,27 @@ def showInteractionsGraph(statistics, **kwargs):
     node_distance = kwargs.pop('node_distance', 5)
     font_size = kwargs.pop('font_size', 14)
     seed = kwargs.pop('seed', 42)
-    cutoff = kwargs.pop('cutoff', 3)
+    cutoff = kwargs.pop('cutoff', 0.1)
+    multiple_chains = kwargs.pop('cutoff', False)
```
jamesmkrieger, line 1493: Should this really take from cutoff?

karolamik13: I am using cutoff to remove those pairs of contacts which occurred only a few times. 0.1 means that it will display pairs that were detected in 10% frames of the trajectory (if that was a single contact between residues).

jamesmkrieger: But it shouldn't be what's used for setting the multiple chains kwarg

karolamik13: Ohh.. you are right. I forgot to replace the name. Thanks.

jamesmkrieger: You're welcome

jamesmkrieger: Shall I fix it?

karolamik13: Yes, please.

jamesmkrieger: done

prody/proteins/interactions.py
```diff
@@ -0,0 +1,3288 @@
...
+        coords = (atoms._getCoords() if hasattr(atoms, '_getCoords') else
+                    atoms.getCoords())
+    except AttributeError:
+        try:
+            checkCoords(coords)
+        except TypeError:
+            raise TypeError('coords must be an object '
+                            'with `getCoords` method')
+
+    distA = kwargs.pop('distA', 3.5)
```
jamesmkrieger, line 311: we can't pop these here if we want to provide them with **kwargs below. We'd have to provide them again separately. This could be useful if you want different default values, but probably we don't.

Shall I just remove them here?

prody/proteins/interactions.py
```diff
@@ -0,0 +1,3288 @@
...
+    
+def calcChHydrogenBonds(atoms, **kwargs):
+    """Finds hydrogen bonds between different chains.
+    See more details in calcHydrogenBonds().
+    
+    :arg atoms: an Atomic object from which residues are selected
+    :type atoms: :class:`.Atomic`
+    
+    :arg distA: non-zero value, maximal distance between donor and acceptor.
+        default is 3.0.
```
jamesmkrieger, line 282: also, do we want to change this to 3.5 to match calcHydrogenBonds

karolamik13: Yes, definitely.

jamesmkrieger: done

prody/proteins/interactions.py
```diff
@@ -578,7 +579,9 @@ def calcPiStacking(atoms, **kwargs):
     aromatic_dic = {'TRP':'noh and not backbone and not name CB NE1 CD1 CG',
                 'PHE':'noh and not backbone and not name CB',
                 'TYR':'noh and not backbone and not name CB and not name OH',
-                'HIS':'noh and not backbone and not name CB'}
+                'HIS':'noh and not backbone and not name CB',
```
jamesmkrieger, line 582: Do we need to handle alternative histidine for salt bridges too? I guess not HSE and HSD because these are single protonated neutral but the other one

karolamik13: I was thinking about that too. HSP is protonated in CHARMM ff. We might remove HSE and HSD. Unless we would like to keep them in case during the simulation we would observe some reorganization and then HSD/HSE will come closer to negatively charged residue. Then the prediction of HIS to be HSE might be wrong. Is it sufficient reason to keep those?

prody/proteins/interactions.py
```diff
@@ -154,13 +154,17 @@ def calcHydrophobicOverlapingAreas(atoms, selection):
     writePDB('selRes_'+atoms.getTitle(), selectedHydrophobicRes)
     hydrophobic_structure = 'selRes_'+atoms.getTitle()+".pdb"
 
-    # Direct path is given for tests - it is not working without it [no idea why]    
     try:
```
jamesmkrieger, line 157: use prody PY3K

prody/proteins/interactions.py
```diff
@@ -154,13 +154,17 @@ def calcHydrophobicOverlapingAreas(atoms, selection):
...
         result = subprocess.run(["./regsurf", full_structure, hydrophobic_structure], 
             capture_output=True, text=True)
     except:
-        result = subprocess.run(["/home/karolamik/ProDy/prody/proteins/regsurf", 
-            full_structure, hydrophobic_structure], capture_output=True, text=True)
+        # Python 2:
+        result = subprocess.call(["./regsurf", full_structure, hydrophobic_structure],
+            capture_output=True, text=True)
+        
+        #result = subprocess.run(["/home/karolamik/ProDy/prody/proteins/regsurf",
```
jamesmkrieger, line 166: delete commented part

prody/proteins/interactions.py
```diff
@@ -282,11 +282,15 @@ def calcVolume(atoms, **kwargs):
         by default is 0.0 to include all the results
     :type sasa_volume: float or int
 
+    :arg split_residues: it will provide values for each residue
+        by default is False
+    :type split_residues: True or False
```
jamesmkrieger, line 287: type should be bool

prody/proteins/interactions.py
```diff
@@ -282,11 +282,15 @@ def calcVolume(atoms, **kwargs):
...
+
     :arg resnames: residues name included
-        by default is 'sum'
+        by default is False
         True - will give residue names and values for each residue
         False - will give only the values for each residues
-    :type resnames: True, False, 'sum'    
+    :type resnames: True or False
```
jamesmkrieger, line 293: type should be bool

prody/proteins/interactions.py
```diff
@@ -0,0 +1,3563 @@
...
+    :type selection: str
+    
+    :arg hpb_cutoff: cutoff for hydrophobic overlaping area values
+        by default is 0.0
+    :type hpb_cutoff: float or int
+    
+    :arg cumulative_values: sum of results for pairs of residues or
+        for single residues without distinguishing pairs,
+        by default is None
+    :type cumulative_values: 'pairs' or 'single_residues'
```
jamesmkrieger, line 172: this should be a proper type. Is it a string or a list or what? You can describe what's inside it in the part above

PR #1634, opened 2023-01-14
Fixes to mmcif parsing

odcambc:
The mmCIF parser was failing to handle the CIF files that pymol was exporting. This tweaks the parser to account for those cases.

1. The doneAtomBlock loop now checks if the atom block continues to the concluding line of the file.
2. The parser now checks whether either 'auth_atom_id' or 'label_atom_id' are present
3. Same, but with 'auth_comp_id' and 'label_comp_id
4. Also with 'auth_seq_id' and 'label_seq_id'


jamesmkrieger:
Thanks very much!

jamesmkrieger (approved):

Issue #1637, opened 2023-01-19
filtering Dali record should be by stringency.


odcambc:
Calling filter on a DaliRecord object behaves inconsistently with respect to the criteria. It returns the record's list of PDBs by removing any record that matches _any_ of the cutoffs.

I believe it makes most sense if, for each criteria, it removes matches that are less similar, given that metric, and retains the more similar. Currently this is inconsistent:

length of aligned residues < cutoff, filtered
RMSD < cutoff, filtered
Z score < cutoff, filtered
Identity > cutoff, filtered

cutoff_rmsd and cutoff_identity behave the opposite of what I would expect, where decreasing the RMSD cutoff _decreases_ the number of filtered matches (similarly with identity).

I propose either those two should be switched so that filtering has the same sense. Alternatively, add a flag to change the behavior to filter by stringency.

jamesmkrieger:
I think there was a logic to the old behaviour, probably something like giving a greater versatility to the filtering and allowing the user to reduce redundancy. Your way also definitely makes sense and I appreciate your fix that maintains the default behaviour so issue closed.

We incidentally have a function refineEnsemble that allows RMSD filtering later too and can be used in both directions. It could be nice to have an option for sequence identity filtering like that too. I'll make another issue asking for that.

PR #1638, opened 2023-01-19
Addresses inconsistent cutoffs in filtering

odcambc:
Proposed changes to address #1637 

Adds a flag to filter that makes cutoffs behave consistently. Defaults maintain old behavior.


jamesmkrieger (approved):
Looks good to me

jamesmkrieger:
Adding new features that are found to be helpful while maintaining default behaviour is always welcome. Thank you

PR #1657, opened 2023-02-08
avoid errors from writeNMD atoms

jamesmkrieger:
An alternative could be to check the type of atoms


jamesmkrieger:
> An alternative could be to check the type of atoms

done instead now

jamesmkrieger:
actually, I guess we can support Ensemble objects too, so hasattr would be better

jamesmkrieger:
In the end, we get an attribute error and it's the same thing

PR #1678, opened 2023-03-26
clustenm title to str

jamesmkrieger:
The current implementation let's the title have any time including AtomGroup and makes problems for loadEnsemble:
```
In [14]: ens = prody.loadEnsemble("/home/jkrieger/ScipionUserData/projects/D614G_prody/Runs/003495_ProDyClustENM/clustenm.ens.npz")
ValueError                                Traceback (most recent call last)
Cell In [14], line 1
----> 1 ens = prody.loadEnsemble("/home/jkrieger/ScipionUserData/projects/D614G_prody/Runs/003495_ProDyClustENM/clustenm.ens.npz")

File /losdato/jkrieger/scipion3/software/em/ProDy/prody/ensemble/functions.py:121, in loadEnsemble(filename, **kwargs)
    118         title = None
    120 if isinstance(title, np.ndarray):
--> 121     title = title.item()
    123 if not isinstance(title, str) and title is not None:
    124     try:

ValueError: can only convert an array of size 1 to a Python scalar

In [15]: kwargs = {}

In [16]: filename = "/home/jkrieger/ScipionUserData/projects/D614G_prody/Runs/003495_ProDyClustENM/clustenm.ens.npz"

In [17]: import numpy as np

    ...:         kwargs['allow_pickle'] = True
    ...: 
    ...:     attr_dict = np.load(filename, **kwargs)
    ...:     if '_weights' in attr_dict:
    ...:         weights = attr_dict['_weights']
    ...:     else:
    ...:         weights = None
    ...: 
    ...:     # backward compatibility
    ...:     try:
    ...:         type_ = attr_dict['_type']
    ...:     except KeyError:
    ...:         if weights is not None and weights.ndim == 3:
    ...:             type_ = 'PDBEnsemble'
    ...:         else:
    ...:             type_ = 'Ensemble'
    ...: 
    ...:     try:
    ...:         title = attr_dict['_title']
    ...:     except KeyError:
    ...:         try:
    ...:             title = attr_dict['_name']
    ...:         except KeyError:
    ...:             title = None
    ...: 

In [19]: title
Out[19]: 
array([<Atom: N from 7KEC_1|Chains.B99990001_fixed (index 0)>,
       <Atom: H from 7KEC_1|Chains.B99990001_fixed (index 1)>,
       <Atom: H2 from 7KEC_1|Chains.B99990001_fixed (index 2)>, ...,
       <Atom: C from 7KEC_1|Chains.B99990001_fixed (index 51701)>,
       <Atom: O from 7KEC_1|Chains.B99990001_fixed (index 51702)>,
       <Atom: OXT from 7KEC_1|Chains.B99990001_fixed (index 51703)>],
      dtype=object)
```

 I've taken the title setting line from class Ensemble instead, which ensures that we get a short string. Is there any reason not to do this or, in fact, any reason not to just use the title setting from Ensemble directly?

### DIFF
--- a/prody/dynamics/clustenm.py
+++ b/prody/dynamics/clustenm.py
@@ -112,7 +112,7 @@ class ClustENM(Ensemble):
         self._tmdk = 10.
 
         super(ClustENM, self).__init__('Unknown')   # dummy title; will be replaced in the next line
-        self._title = title
+        self._title = str(title).strip()
 
     def __getitem__(self, index):
 


kaynakb:
There are couple of reasons. When ProDy parses the PDB file, it either assigns the name of the pdb as the title of the AtomGroup or it concatenates the ChainIds to the name thereof. If the user does not give a name, then all of the output files consistently get the same title, e.g.,

4akeA_clustenm.dcd
4akeA_clustenm.ens.npz
4akeA_clustenm_parameters.txt
4akeA_fixed.pdb

It can also create 4akeA directory if the conformers are wanted to be written as separate files.

Moreover, if the user runs multiple ClustENMD simulations for the different chains of the same structure in the same folder, there won’t be any override. 

ClustENMD have been used for proteins with multiple chains. We have not encountered the problem you had. What does getTitle() function returns when you parse your PDB? I expect it to be a string not an array.

jamesmkrieger:
Thanks. That sounds great. 

Wouldn't the same be true if we converted the AtomGroup to str to get its title like other ProDy ensembles?

I had just provided the AtomGroup itself and ClustENM getTitle was giving me an AtomGroup back. I've changed my code to provide a string instead of the AtomGroup, but perhaps there could be a better warning or error.

kaynakb:
By the way, the following is better to create a string instead of str(title).strip() if your title contains an iterable. Assume l is a container, Python or NumPy array, then <some string ' ' or '.'>.join([str(x) for x in l]) if each x is other than a string.

jamesmkrieger:
> By the way, the following is better to create a string instead of str(title).strip() if your title contains an iterable. Assume l is a container, Python or NumPy array, then <some string ' ' or '.'>.join([str(x) for x in l]) if each x is other than a string.

That's exactly what I didn't want. saveEnsemble unpacked the AtomGroup as an iterable of atoms, but I don't want the str from each of the atoms in the AtomGroup any more than I want the atoms themselves.

kaynakb:
I did not suggest it as a solution to your problem since you already
changed your code. str(title).strip() is ambiguous because it does not
reflect your intention due to the fact that the type of title is not
defined.

jamesmkrieger:
Fair enough

PR #1697, opened 2023-05-23
Add membrane exANM and exGNM to apps

jamesmkrieger:


jamesmkrieger:
This is now fixed, so it's ready to go

PR #1699, opened 2023-05-29
no adjustText for many structs

jamesmkrieger:
With lots of structures, adjustText is slow and doesn't really work


jamesmkrieger:
Actually, it would be better to have an option to just turn off adjust text

PR #1706, opened 2023-06-13
Show matrix markersize scatter option

jamesmkrieger:
This allows matrices to be plotted with different size dots by using scatter with markersize. Default behaviour when markersize is not set remains the same


jamesmkrieger:
Default:
```
showContactMap(atoms2.ca)
```
default_5kqm_showContactMatrix

Smaller points to 0.1, adjusting back norm:
```
showContactMap(atoms2.ca, markersize=0.1, norm=plt.Normalize(0,1))
```
0p1_5kqm_showContactMatrix

Bigger points to 10 likewise:
```
showContactMap(atoms2.ca, markersize=10, norm=plt.Normalize(0,1))
```
10_5kqm_showContactMatrix

PR #1707, opened 2023-06-13
actually use vcenter in showMatrix

jamesmkrieger:
Before, it was set to 0. instead

### DIFF
--- a/prody/utilities/catchall.py
+++ b/prody/utilities/catchall.py
@@ -655,7 +655,7 @@ def showMatrix(matrix, x_array=None, y_array=None, **kwargs):
             except ImportError:
                 from matplotlib.colors import TwoSlopeNorm as DivergingNorm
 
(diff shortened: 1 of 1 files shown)


jamesmkrieger:
Here's the result with a random matrix.

```
In [1]: import matplotlib.pyplot as plt
   ...: plt.ion()
   ...: from prody import *
   ...: import numpy as np

In [2]: matrix = np.random.randint(-5,5,(10,10))

In [3]: writeArray('m5_p5_10x10_random.txt', matrix)
Out[3]: '5x5_random.txt'
```

Default:
```
showMatrix(matrix)
```
Figure_4


Old behaviour for vcenter 2 (no change):
```
showMatrix(matrix, vcenter=2)
```
no_change_showMatrix

New behaviour:
```
showMatrix(matrix, vcenter=2)
```
vcen2_showMatrix

Issue #1722, opened 2023-08-03
clustenm app


jamesmkrieger:
This would be helpful for running clustenm more easily outside scripts etc.

PR #1726, opened 2023-08-09
Clustenm app

jamesmkrieger:
This app can run ClustENM and save the outputs in pdb and ens.npz files. 

Example command:
```
prody clustenm Runs/000002_ProDySelect/4ake_atoms.pdb --ngens 2 --number-of-modes 3 --nconfs 5 --rmsd 1. -c 15.0 -g 1.0 --maxclust "(2, 3)" --threshold "None" --solvent imp --force_field None --ionicStrength 0.0 --padding 1.0 --temp 303.15 --t_steps_i 1000 --t_steps_g 7500 --tolerance 10.0 --maxIterations 0 -P 6 -o Runs/000089_ProDyClustENM/clustenm_1 --file-prefix pdbs --no-sim --turbo --outlier --mzscore 3.5
```

Outputs:
```
(scipion3) jkrieger@whipple ~/ScipionUserData/projects/TestProDyClustENM $ ls -ltrah Runs/000089_ProDyClustENM/clustenm_1/
total 1.2M
drwxr-xr-x 2 jkrieger jkrieger 4.0K Aug  9 16:37 pdbs
-rw-r--r-- 1 jkrieger jkrieger 1.2M Aug  9 16:37 pdbs.ens.npz
drwxr-xr-x 3 jkrieger jkrieger 4.0K Aug  9 16:37 .
drwxr-xr-x 5 jkrieger jkrieger 4.0K Aug  9 16:37 ..
(scipion3) jkrieger@whipple ~/ScipionUserData/projects/TestProDyClustENM $ ls -ltrah Runs/000089_ProDyClustENM/clustenm_1/pdbs
total 1.6M
-rw-r--r-- 1 jkrieger jkrieger 265K Aug  9 16:37 0_0.pdb
-rw-r--r-- 1 jkrieger jkrieger 265K Aug  9 16:37 1_0.pdb
-rw-r--r-- 1 jkrieger jkrieger 265K Aug  9 16:37 1_1.pdb
-rw-r--r-- 1 jkrieger jkrieger 265K Aug  9 16:37 2_0.pdb
-rw-r--r-- 1 jkrieger jkrieger 265K Aug  9 16:37 2_1.pdb
drwxr-xr-x 2 jkrieger jkrieger 4.0K Aug  9 16:37 .
-rw-r--r-- 1 jkrieger jkrieger 265K Aug  9 16:37 2_2.pdb
```

### DIFF
--- a/PKG-INFO
+++ b/PKG-INFO
@@ -1,6 +1,6 @@
 Metadata-Version: 1.1
 Name: ProDy
-Version: 2.4.0
+Version: 2.4.1
 Summary: A Python Package for Protein Dynamics Analysis
 Home-page: http://www.csb.pitt.edu/ProDy
 Author: James Krieger, She Zhang, Hongchun Li, Cihan Kaya, Ahmet Bakan, and others
--- a/docs/apps/prody/prody.txt
+++ b/docs/apps/prody/prody.txt
@@ -1,6 +1,4 @@
-usage: prody [-h] [-c] [-v]
-             {anm,gnm,pca,eda,align,blast,biomol,catdcd,contacts,fetch,select}
-             ...
+usage: prody [-h] [-c] [-v] {anm,gnm,pca,eda,align,blast,biomol,catdcd,contacts,fetch,select,energy,clustenm} ...
 
 ProDy: A Python Package for Protein Dynamics Analysis
 
@@ -10,7 +8,7 @@ optional arguments:
   -v, --version         print ProDy version and exit
 
 subcommands:
-  {anm,gnm,pca,eda,align,blast,biomol,catdcd,contacts,fetch,select}
+  {anm,gnm,pca,eda,align,blast,biomol,catdcd,contacts,fetch,select,energy,clustenm}
     anm                 perform anisotropic network model calculations
     gnm                 perform Gaussian network model calculations
     pca                 perform principal component analysis calculations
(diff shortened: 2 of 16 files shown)


jamesmkrieger:
I need to remove thread pool limits because they don’t work, probably they’re applying limits on function calls and not the downstream functions where it matters

kaynakb:
Hi James,

The following line contradicts with the definition and usage of the outlier parameter. Please see the original doc therefore:

('no_outlier', 'whether to not exclude conformers detected as outliers in each generation when using explicit solvent', False)

The original definition is:

:arg outlier: Exclusion of conformers detected as outliers in each generation.
    Default is True for implicit solvent. Outliers, if any, are detected by
    the modified z-scores of the conformers' potential energies over a generation.
    Note: It is automatically set to False when explicit solvent model is being used.
:type outlier: bool

As it is pointed out in the Note above, if the solvent model is chosen as an explicit one, then this parameter is gonna be set to False. The choice whether to use it or not is available when the solvent model is implicit.

Thank you,

Burak

jamesmkrieger:
Thanks. Good catch. It should say implicit solvent then, not explicit.

The rest is on purpose. We can only use booleans in the apps to set something to true, so if we want to change from a default of true then we set the negative to true instead

jamesmkrieger:
Addresses #1722

PR #1727, opened 2023-08-10
Water bridging module

fdoljanin:
New module for detection and analysis of water bridges in proteins.

Water bridges are hydrogen bonds formed when water molecules bridge between two or more protein residues, creating a hydrogen-bonding network stabilising protein-protein and protein-ligand interactions. The developed code aims to identify and analyze conserved water molecules and protein regions involved in interactions mediated by water bridges. This will benefit the scientific community in understanding protein interactions and stability.

# Methods
More information on the methodology can be found [here](https://drive.google.com/file/d/1CmD1F66DqQwt4_PKWM5fq-NWRZtWCYFf/view?usp=sharing).

`calcWaterBridges` returns calculated bridges in one of three output types:
1. indices -bridge is just a list of indices of atoms involved in a bridge
2. atomic - bridge is an object which has _waters_ and _proteins_ attributes, and each is a list of Atom objects involved in bridging
3. info - bridge is a list of strings containing bridge info

`calcWaterBridgesTrajectory` does the same but for multiple frame trajectory
`getWaterBridgesInfoOutput` converts atomic output from calcWaterBridges/Trajectory into info
`calcWaterBridgesStatistics` returns statistics for trajectory; analyses every two residues involved in the bridge (number of occurrences, average distance, distance std)
`getWaterBridgeStatInfo` converts indices output from calcWaterBridgesStatistics to info
`calcWaterBridgeMatrix` calculates matrix based on calcWaterBridgesStatistics indices output
`showWaterBridgeMatrix` shows a matrix based on calcWaterBridgesStatistics indices output
`calcBridgingResiduesHistogram` shows a histogram with frame occurrence count for each residue
`calcWaterBridgesDistribution` shows distributions for various metrics
`savePDBWaterBridges` saves one PDB with occupancy = 1 for involved residues and just waters that participate in the bridge
`savePDBWaterBridgesTrajectory` is the same as savePDBWaterBridges but for multiple frames + modifies beta so that it shows the percentage of interactions for a residue
`saveWaterBridges` saves atomic water bridges into an info file for read-only, or atomic file so that it can be loaded faster next time instead of computing again
`parseWaterBridges` parses saveWaterBridges output so that it behaves like it was returned from calcWaterBridges/Trajectory

This module was developed with mentorship from @karolamik13 and is part of InSty development project.

### DIFF
--- a/prody/proteins/waterbridges.py
+++ b/prody/proteins/waterbridges.py
@@ -10,7 +10,6 @@ __email__ = ['karolamik@fizyka.umk.pl', 'fdoljanin@pmfst.hr']
 
 from collections import deque
 from enum import Enum, auto
-from numpy import *
 
 from prody import LOGGER, Atom
 from prody.measure import calcAngle
@@ -69,6 +68,12 @@ class RelationList:
         self.nodes[key] = value
 
 
+class HydrogenBond:
+    def __init__(self, donor, acceptor):
+        self.donor = donor
+        self.acceptor = acceptor
+
+
 def getSameResidueAtoms(observedAtom, atoms):
     return atoms.select(f'resnum {observedAtom.getResnum()}')
 
@@ -89,8 +94,8 @@ def getNeighborsForAtom(observedAtom, atoms, dist):
 def getRelationsList(hydrogenBonds, numAtoms):
     relations = RelationList(numAtoms)
     for bond in hydrogenBonds:
-        atomNode_0 = relations.createIfNotExists(bond[0])
-        atomNode_1 = relations.createIfNotExists(bond[1])
+        atomNode_0 = relations.createIfNotExists(bond.donor)
+        atomNode_1 = relations.createIfNotExists(bond.acceptor)
 
         atomNode_0.bonds.append(bond)
         atomNode_1.bonds.append(bond)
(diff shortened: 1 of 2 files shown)


prody/proteins/waterbridges.py
```diff
@@ -0,0 +1,1045 @@
...
+
+"""This module detects, predicts and analyzes water bridges.
+"""
+
+__author__ = 'Karolina Mikulska-Ruminska'
+__credits__ = ['Frane Doljanin', 'Karolina Mikulska-Ruminska']
+__email__ = ['karolamik@fizyka.umk.pl', 'fdoljanin@pmfst.hr']
+
+import numpy as np
+import matplotlib.pyplot as plt
```
jamesmkrieger, line 11: This needs to be imported in functions not generally, so matplotlib is not a hard a dependency of ProDy

prody/proteins/waterbridges.py
```diff
@@ -0,0 +1,1045 @@
...
+        self.waters = waters
+
+
+def getInfoOutput(waterBridgesAtomic):
+    output = []
+    for bridge in waterBridgesAtomic:
+        bridgeOutput = []
+
+        for atom in bridge.proteins:
+            residueInfo = f"{atom.getResname()}{atom.getResnum()}"
```
jamesmkrieger, line 260: we can't use fstrings and keep compatibility with python 2.7

jamesmkrieger: I'll see if I can do try/except for importing this module

PR #1728, opened 2023-08-10
cif fixes for more discrepancies in the format

jamesmkrieger:
Not all mmCIF files follow as strict rules as we'd like


jamesmkrieger:
Thank you

PR #1730, opened 2023-08-10
fix TER lines

jamesmkrieger:
This is the format that TER lines actually have in pdb files (https://www.wwpdb.org/documentation/file-format-content/format33/sect9.html#TER) and also the END lines (https://www.wwpdb.org/documentation/file-format-content/format33/sect11.html#END)

Modeller was complaining about these before


jamesmkrieger:
This needs considerable fixing still

PR #1752, opened 2023-09-09
MMTF Parser

vratins:
Added a basic parser that can read in either an mmtf file or a mmtf structure (decoder) object into an atom group. Includes header retrieval for biomolecule building.

### DIFF
--- a/prody/proteins/mmtffile.py
+++ b/prody/proteins/mmtffile.py
@@ -1,3 +1,8 @@
+# -*- coding: utf-8 -*-
+"""This module defines functions for parsing MMTF files and structure objects.
+
+.. MMTF files: https://mmtf.rcsb.org"""
+
 from numbers import Number
 import os.path
 from collections import OrderedDict
@@ -18,7 +23,6 @@ try:
 except:
     print("Install mmtf to read in mmtf structure objects (e.g. pip install mmtf-python)")
 
-#function to parse mmtf files or mmtf structure objects
 def parseMMTF(mmtf_struc, **kwargs):
 
     chain = kwargs.pop('chain', None)
--- a/prody/proteins/mmtffile.py
+++ b/prody/proteins/mmtffile.py
@@ -1,0 +1,256 @@
+from numbers import Number
+import os.path
+from collections import OrderedDict
+
+from prody.atomic import AtomGroup
+from prody.atomic import flags
+from prody.atomic import ATOMIC_FIELDS
+
+from prody.utilities import openFile, isListLike, copy
+from prody import LOGGER
+
+import struct as st
+import numpy as np
+
+try:
+    from mmtf import fetch, parse
+    import mmtf
+except:
+    print("Install mmtf to read in mmtf structure objects (e.g. pip install mmtf-python)")
+
+#function to parse mmtf files or mmtf structure objects
+def parseMMTF(mmtf_struc, **kwargs):
+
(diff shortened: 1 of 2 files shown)


jamesmkrieger:
Thanks. Can you add a test for this too, please

setup.py
```diff
@@ -14,7 +14,7 @@
 if sys.version_info[:2] == (2, 7) or sys.version_info[:2] <= (3, 5):
     INSTALL_REQUIRES=['numpy>=1.10,<1.24', 'biopython<=1.76', 'pyparsing', 'scipy']
 else:
-    INSTALL_REQUIRES=['numpy>=1.10,<1.24', 'biopython<=1.79', 'pyparsing', 'scipy', 'setuptools']
+    INSTALL_REQUIRES=['numpy>=1.10,<1.24', 'biopython<=1.79', 'pyparsing', 'scipy', 'setuptools', 'mmtf-python']
```
jamesmkrieger, line 17: You probably don’t need to add another hard dependency as you’re using try/except to import it

prody/proteins/mmtffile.py
```diff
@@ -0,0 +1,260 @@
...
+from prody.utilities import openFile, isListLike, copy
+from prody import LOGGER
+
+import struct as st
+import numpy as np
+
+try:
+    from mmtf import fetch, parse
+    import mmtf
+except:
```
jamesmkrieger, line 23: Please specify except ImportError

vratins:
Test and fixes added.

setup.py
dkoes: So where should this dependency be put so that the tests can pass? In `.github/workflows/main.yml`?

jamesmkrieger:
ok, it's working now. I would prefer a pip install in main.yml but I guess it's not that heavy a dependency so it's fine. 

The last thing then is to give the user-facing function(s) a doc string.

Thanks again for contributing this!

jamesmkrieger:
No, I think there are still a few last things to fix:

1. The try/except for the import should be inside the function, so users don’t get the warning every time they import prody

2. There are some redundancies between the two parts of the doc string that need to be cleaned up


Besides that, yes, it sounds good

vratins:
Could you let me know what further formatting you'd like to the docstring? I'll make the changes asap.

jamesmkrieger:
Doc string is fine.

the last thing is for the try/except to raise ImportError

jamesmkrieger:
Looks good now. Thanks again!

and thanks, David, for other fixes too!

jamesmkrieger (approved):

prody/proteins/mmtffile.py
```diff
@@ -0,0 +1,360 @@
...
+    :type mmtf_struc: str or MMTF Structure object
+
+    :arg chain: Chain identifier(s) to parse (e.g., 'A' for chain A). If not provided,
+        all chains are parsed. If a PDB ID is used, chain can also be specified as part
+        of the ID (e.g., '1abcA' to parse chain A).
+    :type chain: str, optional
+
+    :arg title: Title to assign to the resulting AtomGroup. If not provided, the
+        title is extracted from the MMTF structure or set to the PDB ID.
+    :type title: str, optional
```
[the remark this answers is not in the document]
jamesmkrieger: Ideally this should change but it’s ok really

PR #1757, opened 2023-10-04
Make sure PDB chain id is only a single character.

dkoes:
Some structures can have multi-character chain identifiers (see 6Z4H). Really Bad Things happen when trying to write these out.

### DIFF
--- a/prody/proteins/pdbfile.py
+++ b/prody/proteins/pdbfile.py
@@ -1525,7 +1525,7 @@ def writePDBStream(stream, atoms, csets=None, **kwargs):
 
             write(pdbline % (hetero[i], serial,
                              atomnames[i], altlocs[i],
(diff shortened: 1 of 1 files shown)


jamesmkrieger:
Have you seen https://github.com/prody/ProDy/pull/1747? Do you think that works too?

dkoes:
#1747 is only for reading files.  This fix is for writing.  You can't parse the 6Z4H PDB file because it doesn't exist - it is only available as mmCIF or mmtf.  If you read it in as an MMTF, the chain ids are set to a three character string and the string formatting to write out the chain id can't handle this, resulting in the rest of the pdb line being displaced and values in the wrong fields.

jamesmkrieger:
ah, ok. Thanks

PR #1758, opened 2023-10-04
Less ambiguous selection when building biomolecule

dkoes:
If a protein has a chain 'x' and a chain '2' and the x happens to be
next to the 2 in the selection string (chain a b c x 2) the parser fails
to realize that x and 2 are chain identifiers.  This makes sure there is
no ambiguity.

There is still a problem with PDBs that have a chain 'bb' (e.g. 2VVJ),
but I'm not up for diving into the parser to fix that.


dkoes:
Any chance of a merge @jamesmkrieger ?

jamesmkrieger:
ok. I'm not sure why it works, but I've made a pdb file with a chain x and it works, so yes.

PR #1764, opened 2023-10-08
Interactions with faster tests

jamesmkrieger:
I think this is ready to merge again


jamesmkrieger:
Actually, I'll do it together with removing python 3.7

PR #1765, opened 2023-10-09
Interactions (hpb improvements)

karolamik13:
Improvements for hpb.so (description, ImportError, different Python versions for hpb.so, docx)


prody/proteins/interactions.py
```diff
@@ -182,7 +181,7 @@ def calcHydrophobicOverlapingAreas(atoms, **kwargs):
             import hpb
             imported_hpb = True
         except ImportError:
-            LOGGER.info('Please provide hpb.so file into the directory.')
+            raise ImportError('Please provide hpb.so file into the directory.')
```
jamesmkrieger, line 184: Doesn’t need to be in directory, could be compiled inside prody

karolamik13: You are right. I will change that, and I will add the description in the main ProDy README.

PR #1771, opened 2023-10-12
fix get unobserved seq

jamesmkrieger:
Without this fix, we can't even parse some mmcif files like 2VVJ at all


jamesmkrieger:
This should also have some tests

jamesmkrieger:
This also has 
- swap of auth_asym_id and label_asym_id 
- trimming away semicolons on the edges of mmcif entries

Probably the trimming needs to be done at a deeper level

jamesmkrieger:
Besides some specific improvement for getting better output, the actual fix part of this is the same as the other one, so I'll merge them together into one

PR #1786, opened 2023-11-05
add seqsep to iterNeighbors

### LINKED ISSUE #1430: Excluding contacts from flanking residues
Hi ProDy team - I keep finding myself having to loop through the tuples returned by findNeighbors to look up the atomic indices and find the residue number in my original model, and then exclude any contacts from a given number of flanking residues (depending on the contact radius I have chosen.) This is the slowest part of my analysis, and when I'm going through >30K pdb files it can end up taking days just for this one step. Is it possible to exclude contacts from *n* flanking residues from the contact list as it is being generated? If not, is this a feature you would consider adding?

jamesmkrieger:
closes #1430


jamesmkrieger:
This could probably do with a test

Issue #1790, opened 2023-11-08
6UWI doesn't parse


dkoes:
6UWI is the only entry in the PDB that prody can't parse (whether using parsePDB, parseCIF, or parseMMTF).  Have not investigated it at all but making an issue hoping someone else does.

jamesmkrieger:
The problem here is that the default altloc is 1 instead of A

PR #1792, opened 2023-11-08
adhoc fix for 6uwi for mmCIF

jamesmkrieger:
partially helps with #1790 

We still need a fix for MMTF

### DIFF
--- a/prody/proteins/ciffile.py
+++ b/prody/proteins/ciffile.py
@@ -302,7 +302,6 @@ def _parseMMCIFLines(atomgroup, lines, model, chain, subset,
     doneAtomBlock = False
     start = 0
     stop = 0
-    warnedAltloc = False
     while not doneAtomBlock:
         line = lines[i]
         if line[:11] == '_atom_site.':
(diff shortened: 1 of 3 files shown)


prody/proteins/ciffile.py
```diff
@@ -432,7 +436,12 @@ def _parseMMCIFLines(atomgroup, lines, model, chain, subset,
 
         alt = line.split()[fields['label_alt_id']]
         if alt not in which_altlocs and which_altlocs != 'all':
-            continue
+            try:
+                alt = ascii_uppercase[int(alt)]
```
dkoes, line 440: Does this work? Wouldn't it make the altloc 'B' when the default is 'A'?

jamesmkrieger: Oh, right, true. I need to subtract 1.

dkoes:
But isn't this going to produce warnings in normal cases - whenever you have a non-interger altloc that isn't requested?

jamesmkrieger:
> Not sure this works? What I'd like to to with parseMMTF is revert the requested altloc from A to the first altloc in the file if no atoms are selected.

Good point. We still want to filter them. Thanks

dkoes:
And what if a file has altloc 'A' and altloc '1'? Then this will merge them.

jamesmkrieger:
ok, the altloc masking works for CIF too

It looks more complicated to set it for PDB where things seem to be assigned dynamically through the loop.

jamesmkrieger:
It looks like this breaks the multi-model case

PR #1783, opened 2023-11-02
fix altloc to coordinate sets

jamesmkrieger:
Old behaviour:
```
$ prody select "name CA" 6flr.pdb
@> WARNING failed to parse altloc 'B' at line 1821, could not read coordinates
@> WARNING failed to parse altloc 'B' at line 1829, could not read coordinates
@> WARNING failed to parse altloc 'B' at line 1833, could not read coordinates
@> WARNING failed to parse altloc 'B' at line 1837, could not read coordinates
@> WARNING failed to parse altloc 'B' at line 1841, could not read coordinates
@> WARNING failed to parse altloc 'B' at line 1845, could not read coordinates
@> WARNING failed to parse altloc 'B' at line 10264, could not read coordinates
@> WARNING failed to parse altloc 'B' at line 10272, could not read coordinates
@> WARNING failed to parse altloc 'B' at line 10276, could not read coordinates
@> WARNING failed to parse altloc 'B' at line 10280, could not read coordinates
@> WARNING failed to parse altloc 'B' at line 10284, could not read coordinates
@> WARNING failed to parse altloc 'B' at line 10288, could not read coordinates
@> WARNING failed to parse altloc 'B' at line 10292, could not read coordinates
@> 13 out of 26 altloc 'B' lines were parsed.
@> Altloc 'B' is appended as a coordinate set to atomgroup 6flr.
@> 6073 atoms and 2 coordinate set(s) were parsed in 0.06s.
@> Secondary structures were assigned to 484 residues.
@> Selection 'name CA' matched 741 atoms.
@> Selection is written into: 6flr_selected.pdb
```

Fixed behaviour:
```
$ prody select "name CA" 6flr.pdb
@> 6086 atoms and 1 coordinate set(s) were parsed in 0.05s.
@> Secondary structures were assigned to 484 residues.
@> Selection 'name CA' matched 743 atoms.
@> Selection is written into: 6flr_selected.pdb
```

PR #1793, opened 2023-11-08
More robust handling of altloc

dkoes:
This has a fix for #1790 as well as a fix for pdb output (chains must be single character or the file is broken) that I apparently hadn't made a PR for.


prody/proteins/mmtffile.py
```diff
@@ -221,7 +223,7 @@ def _bio_transform(dec):
         ret[t['name']] = L
     return ret
 
-def set_info(atomgroup, mmtf_data,get_bonds=False,altloc_sel='A'):
+def set_info(atomgroup, mmtf_data,get_bonds=False,altloc_sel=None):
```
jamesmkrieger, line 226: I wouldn't use None here because that has a defined behaviour in the rest of ProDy of handling altlocs in a completely different way of making multiple coordinate sets as I mentioned earlier (see #1783 where I fixed some bugs with that).

I also don't see the point of setting it to None and then resetting it to 'A' if it's None.

jamesmkrieger:
Besides, the one comment, I think this looks good. I'll try and do something similar to mmCIF and PDB.

I also see that this only enables the option of asking for 1 altloc and not a few of them like we have mmCIF and PDB.

prody/proteins/mmtffile.py
jamesmkrieger: You may want to consider keeping all entries with no altloc too. That's what happens for mmCIF if which_altloc=None:
https://github.com/prody/ProDy/blob/cdbed98fb33d6073bc23491549c54c7c5fb87db9/prody/proteins/ciffile.py#L364

prody/proteins/mmtffile.py
```diff
@@ -322,7 +324,11 @@ def set_info(atomgroup, mmtf_data,get_bonds=False,altloc_sel='A'):
     mask = np.full(asize, True, dtype=bool)
     if altloc_sel != 'all':
         #mask out any unwanted alternative locations
-        mask = (altlocs == '') | (altlocs == altloc_sel)
+        default_altloc = altloc_sel if altloc_sel != None else 'A'
+        mask = (altlocs == '') | (altlocs == default_altloc)
+        if np.all(mask == False) and altloc_sel == None and len(altlocs):
+            #nothing selected, use first altloc; 6uwi
+            mask = altlocs == altlocs[0]
```
jamesmkrieger, line 331: mask = (altlocs == '') | (altlocs == altlocs[0])

jamesmkrieger (changes requested):

prody/proteins/mmtffile.py
dkoes: If you get here, the mask was all false so you have no altlocs == ''

prody/proteins/mmtffile.py
dkoes: The point of setting it to None is to indicate that it was not set by the user, which I can't tell if it is set to A by default.  The idea being if it is one of the 60 PDBs where every atom has an altloc specified and the user asks for 'A' and there is no 'A' (which is only true for 6uwi) they should get an empty molecule back.

prody/proteins/mmtffile.py
jamesmkrieger: Fair enough

prody/proteins/mmtffile.py
jamesmkrieger: Good point 

The case where None gave multiple coordinate sets was surprising anyway

jamesmkrieger (approved):

PR #1796, opened 2023-11-08
New features in calcSminaBindingAffinity

karolamik13:
I added atom_terms parameter, more information can be retrieved from SMINA, and I fixed stop_frame in multi-model PDB.
The code of calcSminaBindingAffinity is now more compact after improvements.

### DIFF
--- a/prody/proteins/interactions.py
+++ b/prody/proteins/interactions.py
@@ -2380,6 +2380,11 @@ def calcSminaBindingAffinity(atoms, trajectory=None, **kwargs):
                            by default is "vina"
     
     :type ligand_selection: str
+    
+    :arg atom_terms: write per-atom interaction term values
+                     by default is False,
+    :type atom_terms: bool
+     
 
     SMINA installation is required to compute ligand binding affinity:
     >> conda install -c conda-forge smina       (for Anaconda)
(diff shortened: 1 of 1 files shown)


prody/proteins/interactions.py
```diff
@@ -2358,9 +2359,48 @@ def showLigandInteraction_VMD(atoms, interactions, **kwargs):
...
+                weight = float(parts[0])
+                term = ' '.join(parts[1:])
+                term_dict[term] = weight
+        data.update(term_dict)
+        
+    term_values_match = re.search(r'Term values, before weighting:\n##\s+(.*?)\n', result, re.DOTALL)
+    if term_values_match:
+        term_values_text = term_values_match.group(1)
+        term_values_array = np.array([float(value) for value in term_values_text.split()])
+        data['Term values, before weighting'] = term_values_array.tolist()
```
jamesmkrieger, line 2396: You could maybe use the term_values_text for the keys here

karolamik13: Why exactly? With the current code, we have access to all that we need.

jamesmkrieger: I don’t remember what I meant. I think I was looking for a way of accessing the keys and values together rather than separately. 

I’ll have another look tomorrow

PR #1817, opened 2024-01-18
update docs to include SignDy, InSty and WatFinder names

jamesmkrieger:
Updating the rst files isn't super helpful if we can't update the websites but it's a start.


jamesmkrieger:
For SignDy, I've just made it so we can see this:
```
In [1]: from prody import *

In [2]: dynamics.signature?
Type:        module
String form: <module 'prody.dynamics.signature' from '/home/jkrieger/software/scipion3/software/em/ProDy/prody/dynamics/signature.py'>
File:        ~/software/scipion3/software/em/ProDy/prody/dynamics/signature.py
Docstring:  
This module defines functions for signature dynamics (SignDy), analyzing normal modes obtained 
for conformations in an ensemble.
```

If people do a search of the prody repo for signdy, this should also show up

Issue #1840, opened 2024-03-12
Failure to correctly parse all ligand chains in complex PDB file


amorehead:
Hello.

It seems that with Prody version 2.4.1 (same behavior with the latest changes in `master`) not all unique ligand chains can be parsed for the attached complex (protein-ligand-nucleic acid) PDB file using the following code.
```bash
pdb = parsePDB(complex_filepath)
protein = pdb.select("protein")
ligand = pdb.select("not (protein or nucleic or water)")
```
In particular, chain `M` amongst the ligand chains is ignored by this parsing and selection logic. Would you happen to know what might be causing this parsing failure? I've tried viewing `ligand.getChids()` as well as `list(ligand.getHierView().iterChains())`, but chain `M` always seems to be missing. I also tried calling `ligand.setSegnames(ligand.getChids())`, but this didn't solve the issue either. Might this be related to some kind of misalignment between segment IDs and chain IDs (even though I can't spot such a misalignment)?

[Complex.pdb.txt](https://github.com/prody/ProDy/files/14579373/Complex.pdb.txt)

amorehead:
It's definitely a PDB file. This `complex.pdb` file corresponds to the protein-ligand-nucleic acid complex structure which is denoted as target H1171v1 in CASP15 (and corresponds to PDB ID: 7PBL (https://www.rcsb.org/structure/7pbl)).

jamesmkrieger:
ok, thanks

amorehead:
Note that, using BioPython, I was able to parse all of the six ligand chains successfully. With ProDy, it's possible that chain `M` having a different residue name from all the other ligand chain's residue names is somehow throwing off ProDy.

jamesmkrieger:
Can you provide the file please? I don't get a chain M when parsing with the pdb id and displaying the pdb file on the rcsb pdb doesn't show one either.

jamesmkrieger:
I also don't get a chain M when using Biopython from the file downloaded from the pdb by prody:
```
In [15]: ag = parsePDB('7PBL', compressed=False)

In [16]: struct = PDBParser().get_structure('7pbl', file='7pbl.pdb')
/home/jkrieger/software/miniconda/envs/scipion3/lib/python3.8/site-packages/Bio/PDB/StructureBuilder.py:89: PDBConstructionWarning: WARNING: Chain A is discontinuous at line 32028.
  warnings.warn(
[...]

PR #1655, opened 2023-02-05
remove np alias types

jamesmkrieger:
These are deprecated and get removed in numpy 1.24, so now we can hopefully support it as requested in #1594

PR #1842, opened 2024-03-14
More nucleosides

jamesmkrieger:
Inspired by #1840

This extends the nucleosides selection macro to include the nucleosides themselves ADN, CTN, THM and URI (guanosine was already there as GMP) as well as the following and a few others:
- 5GP (guanidine-5'-monophosphate) - not found before as GMP was used
- CMP (cyclic AMP) and C5P (cytidine-5'-monophosphate) - again missed because of unexpected usage
- TPP (thymidine-5'-disphosphate)
- AGS (ATP-gamma-S)


jamesmkrieger:
This now captures ADP and AGS in nucleoside unlike previously

jamesmkrieger:
No idea what makes these tests time out

jamesmkrieger:
> No idea what makes these tests time out

or generally take so long. The same is true for #1817 but not for #1655, even though I updated both.

PR #1849, opened 2024-03-26
Logistic regression for differences between sub-ensembles

jamesmkrieger:
Similar to LDA, but works better


jamesmkrieger:
Something isn't quite working yet

jamesmkrieger:
> Something isn't quite working yet

Now everything does

PR #1853, opened 2024-03-28
Add Bioexcel cv19 database

jamesmkrieger:
We now have functions to fetch, convert (using mdtraj) and parse files from the BioExcel-CV19 simulation database.


jamesmkrieger:
This needs tests too. We're also missing tests for parsePSF and writePSF that should be added too

jamesmkrieger:
We now have tests for helper functions, structure functions and topology functions. We still need them for trajectory ones.

For convert xtc to dcd, we will also need a data file. It may also be good to move this more general function somewhere else, but probably it's fine.

jamesmkrieger:
Try conda-forge mdtraj and find stringio fix

jamesmkrieger:
> Try conda-forge mdtraj and find stringio fix

conda-forge mdtraj was tried before. Python 2 just can't support either version.

jamesmkrieger:
Need to change 2.7 check to use float instead of str too

jamesmkrieger:
And there are still more places to check for seed alignment not full too

jamesmkrieger:
> Try conda-forge mdtraj and find stringio fix

not sure what happened with the stringio thing but it seems ok again

PR #1861, opened 2024-04-09
add exGNM to save load model

jamesmkrieger:


jamesmkrieger:
bioexcel tests fail sporadically. It could be because we're making too many requests at once

PR #1929, opened 2024-08-05
Mymaster

prod-feng:


prody/dynamics/sampling.py
```diff
@@ -49,11 +49,11 @@ def sampleModes(modes, atoms=None, n_confs=1000, rmsd=1.0):
     :math:`[r_1^k r_2^k ... r_m^k]` are normally distributed random numbers
     generated for conformation :math:`k` using :func:`numpy.random.randn`.
 
-    RMSD of the new conformation from :math:`R_0` can be calculated as
+    RMSD of the new TEST conformation from :math:`R_0` can be calculated as
```
jamesmkrieger, line 52: I don't see the word test on http://3.15.148.93/manual/reference/dynamics/sampling.html?highlight=samplemodes#prody.dynamics.sampling.sampleModes

prody/dynamics/sampling.py
```diff
@@ -49,11 +49,11 @@ def sampleModes(modes, atoms=None, n_confs=1000, rmsd=1.0):
...
 
     .. math::
 
-      RMSD^k = \\sqrt{ {\\left( s \\sum_{i=1}^{m} r_i^k \\lambda^{-0.5}_i u_i  \\right)}^{2} / N } = \\frac{s}{ \\sqrt{N}} \\sqrt{ \\sum_{i=1}^{m} (r_i^k)^2 \\lambda^{-1}_i  }
+      RMSD^k = \\sqrt{ { \\left( s \\sum_{i=1}^{m} r_i^k \\lambda^{-0.5}_i u_i  \\right ) }^{2} / N } = \\frac{s}{ \\sqrt{N}} \\sqrt{ \\sum_{i=1}^{m} (r_i^k)^2 \\lambda^{-1}_i  }
```
jamesmkrieger, line 56: We now have square brackets that look reasonable but the ^2 is too low down

docs/release/index.rst
```diff
@@ -9,24 +9,5 @@ Release Notes
...
-   v1.0_series
-   v0.9_series
-   v0.8_series
-   v0.7_series
-   v0.6_series
-   v0.5_series
-   v0.2_series
-   v0.1_series
+
+   ./v*
```
jamesmkrieger, line 13: Does this keep the order starting at the more recent versions? Otherwise you may want to take from my recent commit instead

jamesmkrieger (commented):
I am not approving these yet as there are still some queries

PR #1931, opened 2024-08-07
Parallel savePDBWaterBridgesTrajectory

jamesmkrieger:
From segname PROA from CCT, this gives a speed-up of about 3.3x for 10 frames.

Without the fix the time is 163.88 seconds and with the fix the time is 48.96 seconds.

### DIFF
--- a/docs/release/index.rst
+++ b/docs/release/index.rst
@@ -10,6 +10,10 @@ Release Notes
    :maxdepth: 2
    :glob:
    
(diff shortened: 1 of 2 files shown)


karolamik13 (approved):
I tested the code on caseStudy5.ipynb - everything is working fine. 
I performed the calculations suggested by James - with one chain, it is working fine.

PR #1938, opened 2024-08-14
Remove undoc-members for InSty to get docs shown

jamesmkrieger:


jamesmkrieger:
Having checked through the emails, I can see we don’t need this

PR #1942, opened 2024-08-22
update website throughout prody

jamesmkrieger:
Previously, I only updated the README. Then I realised that PKG-INFO needed to change too and probably these other files too.

There's one left at https://github.com/prody/ProDy/blob/main/docs/devel/website.rst, which I don't know what to do with.

https://github.com/prody/ProDy/blob/main/docs/devel/website.rst?plain=1#L17
First log in to the ProDy webserver (prody.csb.pitt.edu) then run the following::

https://github.com/prody/ProDy/blob/main/docs/devel/website.rst?plain=1#L37
http://prody.csb.pitt.edu/test_prody/_build/html/ in your web browser.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,4 +1,4 @@
-This is a first draft, which is based on the ProDy website page http://prody.csb.pitt.edu/manual/devel/develop.html. Please see that page for updates.
+This is a first draft, which is based on the ProDy website page http://www.bahargroup.org/prody/manual/devel/develop.html. Please see that page for updates.
 
 
 Install Git and a GUI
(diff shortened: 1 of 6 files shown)


prod-feng:
https://github.com/prody/ProDy/blob/main/docs/devel/website.rst,

I suggest to change the pitt url to an example fake url, like

It's recommended to have the symbolic link called test_prody pointing to your build directory instead and then you can monitor changes by going to http://**yourdomai**n/test_prody/_build/html/ in your web browser.

prod-feng:
First log in to your ProDy webserver and then run the following::

jamesmkrieger:
I agree and I see you did it on your version, so this one should be fine as it is. Thanks Feng!

atbogetti (approved):

PR #1943, opened 2024-08-22
fix max_proc 1 multi-model atoms

jamesmkrieger:
This is a critical fix that needs to go into the release. Otherwise, watfinder trajectory analysis wouldn't work for model-model pdb inputs with max_proc == 1.


karolamik13 (approved):
I was testing that when I was checking WatFinder functions, and it was working fine.

jamesmkrieger:
great! Thanks

PR #1948, opened 2024-09-03
InSty - changes in calcStatisticsInteractions 

karolamik13:
I added try/except for the energy parameter because I had a problem analyzing a system with two chains and selecting interactions between them. I think it might be an empty list when there is no interaction at all, and that is the reason.

New changes:
- I modified tabulated_energies.txt to include HSD, HSP and HSE for HIS that are used by NAMD
- I found a bug in get_energy() which was fetching all data except the first line
- a reference to tabulated_energies.txt data is now added


prody/proteins/interactions.py
```diff
@@ -180,21 +179,32 @@ def get_energy(pair, source):
...
+    return data_results
 
 
 def showPairEnergy(data, **kwargs):
     """Return energies when a list of interactions is given. Energies will be added to each pair of residues 
     at the last position in the list. Energy is based on the residue types and not on the distances.
-    The unit of energy is kcal/mol. The energies defined as 'IB_nosolv', 'IB_solv' are taken from XX and 
-    'CS' from YY.
+    The unit of energy is kcal/mol. The energies defined as 'IB_nosolv', 'IB_solv' are taken from [OK98]_ and 
+    'CS' from InSty paper.
```
jamesmkrieger, line 194: maybe mention under preparation or something

prody/proteins/interactions.py
```diff
@@ -1988,12 +1998,20 @@ def calcStatisticsInteractions(data, **kwargs):
...
-            }
+            
+            try:
+                stats[element] = {
+                    "stddev": np.round(np.std(values),6),
+                    "mean": np.round(np.mean(values),6),
+                    "weight": np.round(float(len(values))/len(data), 6),
+                    "energy": get_energy([element.split('-')[0][:3], element.split('-')[1][:3]], energy_list_type)
+                }
+            except:
```
jamesmkrieger, line 2009: we should put a specific type of exception. Otherwise, it can be hard to diagnose if something else happens

prody/proteins/tabulated_energies.txt
```diff
@@ -398,3 +398,120 @@ PRO LYS 0.3	-0.5	0.0
 PRO ARG -0.17	-2.43	0.0
 PRO HIS 0.33	-1.8	0.0
 PRO PRO -0.03	-0.83	0.0
+GLY HSE 0.01	-2.47	0.0
```
jamesmkrieger, line 401: maybe you can have a check to replace HSE or HSD or HSP with HIS before querying this, rather than making it look like we have data specifically for these protonation versions

We should probably also check with Carlos what histidine protonation he did use

jamesmkrieger (changes requested):
just some minor things

prody/proteins/tabulated_energies.txt
karolamik13: You are right, but we will not have HSP, HSE, or HSP data anyway. From Carlos, we will have only one type, and from Ivet paper, too. Maybe I can add a comment that we treat histidines HSP, HSE and HSD as HIS, i.e., without distinguishing differences in protonation.

karolamik13: > maybe you can have a check to replace HSE or HSD or HSP with HIS before querying this, rather than making it look like we have data specifically for these protonation versions

I can do it.

karolamik13:
I added a new function - checkNonstandardResidues(), which checks whether we have some non-standard names in PDB. Tomorrow, I will add checks with this function in different places. When we will have non-standard residues the user will be informed. It might be useful to detect multiple side chain conformation ALYS/BLYS in protein structure as well. 

Except that I modified get_energy() and included a dictionary with non-standard names typical for NAMD, AMBER and GROMACS.

prody/proteins/tabulated_energies.txt
jamesmkrieger: > You are right, but we will not have HSP, HSE, or HSP data anyway. From Carlos, we will have only one type, and from Ivet paper, too. Maybe I can add a comment that we treat histidines HSP, HSE and HSD as HIS, i.e., without distinguishing differences in protonation.

Yes, a comment in docs and code would be good

jamesmkrieger:
> I added a new function - checkNonstandardResidues(), which checks whether we have some non-standard names in PDB. Tomorrow, I will add checks with this function in different places. When we will have non-standard residues the user will be informed. It might be useful to detect multiple side chain conformation ALYS/BLYS in protein structure as well.
> 
> Except that I modified get_energy() and included a dictionary with non-standard names typical for NAMD, AMBER and GROMACS.

I don’t think the gromacs ones make sense and I left a comment about that

prody/proteins/interactions.py
```diff
@@ -1988,12 +2079,21 @@ def calcStatisticsInteractions(data, **kwargs):
...
+                LOGGER.warn('energy information is not available for ', element.split('-')[0][:3], element.split('-')[1][:3])
```
jamesmkrieger, line 2091: I agree a warning is a good idea but only if the except is because of that. Otherwise it makes things even harder to figure out

prody/proteins/interactions.py
```diff
@@ -160,6 +160,53 @@ def get_energy(pair, source):
...
+        'HISD': 'HIS',  # GROMACS: protonated at ND1
+        'HISE': 'HIS',  # GROMACS: protonated at NE2
+        'HISP': 'HIS',  # GROMACS: doubly protonated
+
+        # Cysteine (Cys)
+        'CYX': 'CYS',   # Cystine (disulfide bridge)
+        'CYM': 'CYS',   # Deprotonated cysteine, anion (GROMACS)
+
+        # Aspartic acid (Asp)
+        'ASH': 'ASP',   # Deprotonated Asp (GROMACS: ASH, AMBER: AS4)
```
jamesmkrieger, line 180: I think this should say protonated not deprotonated, but please check. ASP and GLU generally have no protons on the side chain carboxylate

jamesmkrieger: Also you have ASH and GLH below so should these be AS4 and GL4?

prody/proteins/interactions.py
karolamik13: I am not sure whether something else might generate a problem here, and if the problem appears, then the user will see a problematic pair of residues.

prody/proteins/interactions.py
karolamik13: For AMBER, those residues are well documented:
https://ambermd.org/tutorials/basic/tutorial7/index.php
or https://emleddin.github.io/comp-chem-website/AMBERguide-protonation.html

NAMD I know well, so I can easily add those, but I cannot find such a list with non-standard names for Gromacs. So it is hard to double-check those. James, you using Gromacs, right? If you could confirm those it would be great.

jamesmkrieger: I don’t usually see non standard residues from gromacs but I can check it

The ions are pretty non standard in gromacs with the latest charmm36 force field though, e.g. CAL for calcium (to not match CA in protein) and CU2P for copper (II) positive. There was less of this with earlier versions

karolamik13: > I don’t usually see non standard residues from gromacs but I can check it
> 
> The ions are pretty non standard in gromacs with the latest charmm36 force field though, e.g. CAL for calcium (to not match CA in protein) and CU2P for copper (II) positive. There was less of this with earlier versions

I just noticed this your previous message with https://manual.gromacs.org/documentation/current/reference-manual/topologies/pdb2gmx-input-files.html#residue-to-building-block-database

We can include those and that is it.

prody/proteins/interactions.py
jamesmkrieger: But only if they analyse the residues or the logs. I don’t know if it’s better to have something fail or make entries with different values that could get missed and/or make something else fail downstream

I guess there’s nothing we can do about it if we’re missing residue pairs. Maybe we can have something in the doc string about not having energies for all pairs with CS energies

prody/proteins/interactions.py
jamesmkrieger: My point with that link was to exactly not include those because they’re only used internally, but I can check it

karolamik13: Ok, but it looks like some of them are the same in AMBER (see links above).
In NAMD those names are in PDB so we will have them after uploading PDB and trajectory into ProDy. I don't know how it is in Gromacs. I will trust you in this case.

jamesmkrieger: It looks like gromacs outputs 4 character names like HISE and HISD so they are usually parsed as HIS but the information is also there if people want it.

karolamik13: Should I change something in the code, then? Do you have any other comments on the code, or can it be accepted in the current version?

jamesmkrieger: I think we should keep the gromacs things like we have then in case people parse long resnames, which I think there was an option for. 

I’ll have another look but I guess we can merge it

karolamik13: Ok. I am making some new improvements to the old InSty code to analyze chains more conveniently, but this should be separate pull request because it may be more radical and should be better tested.

jamesmkrieger: ok, sounds good. you can make another branch for that then

prody/proteins/interactions.py
```diff
@@ -180,21 +229,69 @@ def get_energy(pair, source):
     
     lookup = pair[0]+pair[1]
     
-    return data[np.where(np.array(aa_pairs)==lookup)[0]][0][2:][np.where(np.array(sources)==source)][0]
+    try:
+        data_results = data[np.where(np.array(aa_pairs)==lookup)[0]][0][2:][np.where(np.array(sources)==source)][0]
+    except ImportError:
```
jamesmkrieger, line 234: why is this an import error? There's no import in the try block

prody/proteins/interactions.py
```diff
@@ -180,21 +229,69 @@ def get_energy(pair, source):
...
 
 
 def showPairEnergy(data, **kwargs):
     """Return energies when a list of interactions is given. Energies will be added to each pair of residues 
     at the last position in the list. Energy is based on the residue types and not on the distances.
-    The unit of energy is kcal/mol. The energies defined as 'IB_nosolv', 'IB_solv' are taken from XX and 
-    'CS' from YY.
+    The unit of energy is kcal/mol. The energies defined as 'IB_nosolv' (non-solvent-mediated), 'IB_solv' (solvent-mediated) 
+    are taken from [OK98]_ and 'CS' from InSty paper (under preparation). 
+    Protonation of resiudues is not distinguished. The protonation of residues is not distinguished.
```
jamesmkrieger, line 280: typo: should say `Protonation of residues` not `Protonation of resiudues` (extra u before the d)

prody/proteins/interactions.py
```diff
@@ -1949,8 +2046,12 @@ def showInteractionsGraph(statistics, **kwargs):
 def calcStatisticsInteractions(data, **kwargs):
     """Return the statistics of interactions from PDB Ensemble or trajectory including:
     (1) the weight for each residue pair: corresponds to the number of counts divided by the 
-    number of frames (values >1 are obtained when residue pair creates multiple contacts); 
-    (2) average distance of interactions for each pair [in Ang] and (3) standard deviation [Ang.].
+    number of frames (values >1 are obtained when the residue pair creates multiple contacts); 
+    (2) average distance of interactions for each pair [in Ang], (3) standard deviation [Ang.],
```
jamesmkrieger, line 2050: put (3) on a new line to make it easier to read

jamesmkrieger (changes requested):
still a few suggestions but I think it looks good

I still think it would be better to have the nonstandard residue handling in a different file like prody/atomic/atomic.py where the is AAMAP and MODAAMAP or maybe somewhere in prody/utilities, but I guess it's ok for now

prody/proteins/interactions.py
```diff
@@ -3917,19 +3936,47 @@ def getInteractions(self, **kwargs):
...
             for i in self._interactions_traj:
                 for nr_j,j in enumerate(i):
                     sele_inter.append(filterInteractions(i[nr_j], self._atoms, **kwargs))
+            
+            if replace == True:
+                try:
+                    trajectory = self._traj
+                    numFrames = trajectory._n_csets
+                except:
+                    # If we analyze previously saved PKL file we doesn't have dcd information
```
jamesmkrieger, line 3964: `it doesn't have` not `we doesn't have`

karolamik13: Ok. I will change that. I sent you an e-mail with the Jupyter Notebook file for tests (for a single PDB file) to see how those functions work now.
Now, it is very convenient if someone analyzes chain-chain interactions and wants to select only those interactions.

prody/proteins/interactions.py
karolamik13: Let's wait for Carlos. It might be added. If not, I can add later an additional description.

prody/proteins/interactions.py
karolamik13: I think this is the only possibility. The energy will be taken when the pair of residues exist in the txt file with pairs. I added an additional description in docs.

prody/proteins/interactions.py
jamesmkrieger: Ok, sounds good

prody/proteins/interactions.py
jamesmkrieger: Ok

jamesmkrieger:
> If that is all, James, can you approve the code so we can merge it?

I believe so. I just want to check through it once more

prody/proteins/interactions.py
```diff
@@ -160,17 +161,65 @@ def get_energy(pair, source):
...
+        'SEP': 'SER',   # Phosphorylated serine (GROMACS/AMBER)
+
+        # Threonine (Thr)
+        'TPO': 'THR',   # Phosphorylated threonine (GROMACS/AMBER)
+
+        # Tyrosine (Tyr)
+        'PTR': 'TYR',   # Phosphorylated tyrosine (GROMACS/AMBER)
+
+        # Non-standard names for aspartic and glutamic acids in low pH environments
+        'ASH': 'ASP',   # Protonated Asp
```
jamesmkrieger, line 207: I also confirm that I've seen one of these recently too

prody/proteins/interactions.py
```diff
@@ -160,17 +161,65 @@ def get_energy(pair, source):
...
+        'GLH': 'GLU',   # Protonated Glu
+    }
+    
+    pair = [aa_correction.get(aa, aa) for aa in pair]    
+    
     try:
```
jamesmkrieger, line 213: we could use `if PY3K`, but it's ok

prody/proteins/interactions.py
```diff
@@ -160,17 +161,65 @@ def get_energy(pair, source):
     import numpy as np
     import importlib.resources as pkg_resources
```
jamesmkrieger, line 162: These imports should be handled with `if PY3K` instead then the rest of the code is the same

```
    if PY3K:
        import importlib.resources as pkg_resources 
    else:
         import pkg_resources
```

I can fix this

prody/proteins/interactions.py
jamesmkrieger: actually, it's not - see my comment above for the python 3 import

Also, does insty work for python 2 now?

prody/proteins/interactions.py
```diff
@@ -180,21 +229,69 @@ def get_energy(pair, source):
...
-    return data[np.where(np.array(aa_pairs)==lookup)[0]][0][2:][np.where(np.array(sources)==source)][0]
+    try:
+        data_results = data[np.where(np.array(aa_pairs)==lookup)[0]][0][2:][np.where(np.array(sources)==source)][0]
+    except TypeError:
+        raise TypeError('Please replace non-standard names of residues with standard names.')
+
+    return data_results
+
+
+def checkNonstandardResidues(atoms):
```
jamesmkrieger, line 240: Probably this should return True or False instead of LOGGER outputs. Also, it doesn't seem to be used anywhere

prody/proteins/interactions.py
```diff
@@ -2087,6 +2203,80 @@ def calcDistribution(interactions, residue1, residue2=None, **kwargs):
...
+            (coord1[2] + coord2[2]) / 2
+        ]
+        return midpoint
+
+    all_DUMs = []
+    atoms_ = atoms.copy()
+
+    for i in interactions:
+        if len(i[1].split('_')) <= 3:
+            res1_name = 'chain '+i[2]+' resname '+i[0][:3]+' and resid '+i[0][3:]+' and index '+' '.join(i[1].split('_')[1:])
```
jamesmkrieger, line 2249: I'm surprising we don't need an `and` before `resname` in this line and all the ones below. We don't if it works but it could be easier to read

prody/proteins/interactions.py
```diff
@@ -2838,14 +3028,33 @@ def getInteractions(self, **kwargs):
         :arg selection2: selection string
         :type selection2: str 
         
+        :arg replace: Used with selection criteria to set the new one
+                      If set to True the selection will be replaced by the new one
+        :type replace: True or False
```
jamesmkrieger, line 3033: no. type is bool and default goes above. I can change this

prody/proteins/interactions.py
jamesmkrieger: ok, it's not the same and I'm fixing it now

prody/proteins/interactions.py
```diff
@@ -3377,18 +3586,22 @@ def showFrequentInteractors(self, cutoff=5, **kwargs):
                 y.append(all_y[nr_ii])
 
         if SETTINGS['auto_show']:
-            matplotlib.rcParams['font.size'] = '20' 
-            fig = plt.figure(num=None, figsize=(12,6), facecolor='w')
+            matplotlib.rcParams['font.size'] = '12'
```
jamesmkrieger, line 3589: we should really leave this up to the user

prody/proteins/interactions.py
jamesmkrieger: actually, I don't think this function works at all. I'm reading through it again

jamesmkrieger:
ok, now I approve merging it

jamesmkrieger:
oh, I actually broke something in the utilities

jamesmkrieger:
> oh, I actually broke something in the utilities

actually never mind. I didn't include that in a commit anyway

prody/proteins/interactions.py
```diff
@@ -160,17 +161,62 @@ def get_energy(pair, source):
...
+        # Aspartic acid (Asp)
+        'ASH': 'ASP',   # Protonated Asp
+        'ASPP': 'ASP',
+
+        # Glutamic acid (Glu)
+        'GLH': 'GLU',   # Protonated Glu
+        'GLUP': 'GLU',  # Protonated Glu
+
+        # Lysine (Lys)
+        'LYN': 'LYS',   # Deprotonated lysine (nautral)
```
jamesmkrieger, line 189: actually here's another typo that I'll fix

PR #1965, opened 2024-10-02
InSty improvements

karolamik13:
I fixed a problem with calcStatisticsInteractions(), added some info warnings for InSty tool and WatFinder.
I also added a new function showInteractionsHist(), which generates bar plots with interactions (see below).

sbs_barplot
 
Now we have two options - bar plot and graph like the one below.

sbs_iPla2B

### DIFF
--- a/prody/proteins/interactions.py
+++ b/prody/proteins/interactions.py
@@ -212,12 +212,10 @@ def get_energy(pair, source):
     
     if PY3K:
         with pkg_resources.path('prody.proteins', 'tabulated_energies.txt') as file_path:
-            #print("File with energies: ", file_path) 
             with open(file_path) as f:
                 data = np.loadtxt(f, dtype=str)
     else:
         file_path = pkg_resources.resource_filename('prody.proteins', 'tabulated_energies.txt')
-        #print("File with energies: ", file_path)  
         with open(file_path) as f:
             data = np.loadtxt(f, dtype=str)
     
--- a/prody/proteins/interactions.py
+++ b/prody/proteins/interactions.py
@@ -41,7 +41,7 @@ __all__ = ['calcHydrogenBonds', 'calcChHydrogenBonds', 'calcSaltBridges',
            'calcPiCationTrajectory', 'calcHydrophobicTrajectory', 'calcDisulfideBondsTrajectory',
            'calcProteinInteractions', 'calcStatisticsInteractions', 'calcDistribution',
            'calcSASA', 'calcVolume','compareInteractions', 'showInteractionsGraph',
-           'calcLigandInteractions', 'listLigandInteractions', 
+           'showInteractionsHist', 'calcLigandInteractions', 'listLigandInteractions', 
            'showProteinInteractions_VMD', 'showLigandInteraction_VMD', 
            'calcHydrogenBondsTrajectory', 'calcHydrophobicOverlapingAreas',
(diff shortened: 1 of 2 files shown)


prody/proteins/interactions.py
```diff
@@ -211,12 +211,15 @@ def get_energy(pair, source):
...
+        with pkg_resources.path('prody.proteins', 'tabulated_energies.txt') as file_path:
+            #print("File with energies: ", file_path) 
+            with open(file_path) as f:
+                data = np.loadtxt(f, dtype=str)
     else:
         file_path = pkg_resources.resource_filename('prody.proteins', 'tabulated_energies.txt')
-
-    with open(file_path) as f:
-        data = np.loadtxt(f, dtype=str)
+        #print("File with energies: ", file_path)
```
jamesmkrieger, line 220: the commented line can be deleted

prody/proteins/interactions.py
```diff
@@ -211,12 +211,15 @@ def get_energy(pair, source):
...
+            #print("File with energies: ", file_path) 
+            with open(file_path) as f:
+                data = np.loadtxt(f, dtype=str)
     else:
         file_path = pkg_resources.resource_filename('prody.proteins', 'tabulated_energies.txt')
-
-    with open(file_path) as f:
-        data = np.loadtxt(f, dtype=str)
+        #print("File with energies: ", file_path)  
+        with open(file_path) as f:
```
jamesmkrieger, line 221: these makes sense

prody/proteins/interactions.py
```diff
@@ -2102,7 +2171,7 @@ def calcStatisticsInteractions(data, **kwargs):
                     "energy": get_energy([element.split('-')[0][:3], element.split('-')[1][:3]], energy_list_type)
                 }
             except:
-                LOGGER.warn('energy information is not available for ', element.split('-')[0][:3], element.split('-')[1][:3])
+                LOGGER.warn('energy information is not available for ', element)
```
jamesmkrieger, line 2174: I'll trust this works

prody/proteins/interactions.py
```diff
@@ -2124,6 +2193,10 @@ def calcStatisticsInteractions(data, **kwargs):
         else: pass
     
     statistic.sort(key=lambda x: x[1], reverse=True)
+    
+    if statistic == []:
+        LOGGER.info("Decrease the value of weight_cutoff parameter to obtain the results. Default is 0.2.")
```
jamesmkrieger, line 2198: This could be a bit clearer. Perhaps say that results were filtered out or whatever it is that's happening.

jamesmkrieger (approved):
seems good. I'd just suggest improving the logging when statistic is empty like the error for no waters

prody/proteins/interactions.py
jamesmkrieger: Makes sense but too long for message. How about "No meaningful interactions found. Decrease weight_cutoff to obtain some results (default is 0.2)."?

karolamik13: I replaced it.

jamesmkrieger: Looks good. Ready to merge

prody/proteins/interactions.py
jamesmkrieger: Ok, I agree it’s helpful to have all the details

PR #1967, opened 2024-10-06
WatFinder - findClusterCenters() extension

karolamik13:
I modified findClusterCenters() to include other selections, not only water.

### DIFF
--- a/prody/proteins/waterbridges.py
+++ b/prody/proteins/waterbridges.py
@@ -1337,7 +1337,7 @@ def findClusterCenters(file_pattern, **kwargs):
             sel = coords_all.select(str(selection)+' within '+str(distC)+' of center',
                                 center=coords_all.getCoords()[ii])
        
-        if'water' in selection.split():
+        if 'water' in selection.split() or 'HOH' in selection.split():
             sel = coords_all.select('water within '+str(distC)+' of center', 
                                 center=coords_all.getCoords()[ii])
         
--- a/prody/proteins/waterbridges.py
+++ b/prody/proteins/waterbridges.py
@@ -1334,12 +1334,11 @@ def findClusterCenters(file_pattern, **kwargs):
     removeResid = []
     removeCoords = []
     for ii in range(len(coords_all)):
-        if 'water' not in selection.split():
-            sel = coords_all.select(str(selection)+' within '+str(distC)+' of center',
-                                    center=coords_all.getCoords()[ii])
-       
         if 'water' in selection.split() or np.any([water in selection.split() for water in DEFAULTS['water']]):
             sel = coords_all.select('water within '+str(distC)+' of center', 
+                                    center=coords_all.getCoords()[ii])
+        else:
+            sel = coords_all.select(str(selection)+' within '+str(distC)+' of center',
                                     center=coords_all.getCoords()[ii])
         
         if sel is not None and len(sel) <= int(numC):
--- a/prody/proteins/waterbridges.py
+++ b/prody/proteins/waterbridges.py
@@ -1333,8 +1333,14 @@ def findClusterCenters(file_pattern, **kwargs):
     removeResid = []
     removeCoords = []
     for ii in range(len(coords_all)):
-        sel = coords_all.select('water within '+str(distC)+' of center', 
+        if 'water' not in selection.split():
+            sel = coords_all.select(str(selection)+' within '+str(distC)+' of center',
                                 center=coords_all.getCoords()[ii])
+       
+        if'water' in selection.split():
+            sel = coords_all.select('water within '+str(distC)+' of center', 
+                                center=coords_all.getCoords()[ii])
+        
         if sel is not None and len(sel) <= int(numC):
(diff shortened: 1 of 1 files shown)


karolamik13:
Before you ask James, I included two types of _ifs_ (separate one for water) because I wanted to be sure that the whole water molecule will be considered in exwithin and not only oxygen, and that would happen if the selection would be not water.

jamesmkrieger:
This doesn't work if the selection has water and something else in it. We should either make a loop for that or raise an error if people try it

jamesmkrieger:
Why is HOH added to water selection? Probably it's better to check if any of the types of water are in there

jamesmkrieger:
> Why is HOH added to water selection? Probably it's better to check if any of the types of water are in there

done

jamesmkrieger:
I'm also a bit confused about why we would want to include the whole water if someone specifically puts in a selection that picks specific atoms. Surely it's better to keep using the provided selection and just update the default one?

karolamik13:
HOH is added to the selection because when you download PDBs from Protein Data Bank, and you analyze multiple structures as we did for protein kinase A, you need to change the selection to _resname HOH and name O2_. We have that in WatFinder paper in Jupyter Notebooks. So, there is no "water" in that selection, and we still want to treat HOH as water because it is a typical PDB name. In the code where we do not have "water" in the selection, it will take your selection so it will check _resname HOH and name O2_ without hydrogens, and then the distC will be different than we had before (for WatFinder paper).

karolamik13:
> > Why is HOH added to water selection? Probably it's better to check if any of the types of water are in there
> 
> done

Definitely. Thank you

jamesmkrieger:
> HOH is added to the selection because when you download PDBs from Protein Data Bank, and you analyze multiple structures as we did for protein kinase A, you need to change the selection to _resname HOH and name O2_. We have that in WatFinder paper in Jupyter Notebooks. So, there is no "water" in that selection, and we still want to treat HOH as water because it is a typical PDB name. In the code where we do not have "water" in the selection, it will take your selection so it will check _resname HOH and name O2_ without hydrogens, and then the distC will be different than we had before (for WatFinder paper).

I still don't understand this. HOH should be recognised as water and we should be able to put "water" in the selection string to catch it.

I also don't get why we use O2 or OH2 and then include the whole water molecule. It still feels like something that we should control at the level of the input selection and not through the internal code because it could give users unexpected results.

karolamik13:
> > HOH is added to the selection because when you download PDBs from Protein Data Bank, and you analyze multiple structures as we did for protein kinase A, you need to change the selection to _resname HOH and name O2_. We have that in WatFinder paper in Jupyter Notebooks. So, there is no "water" in that selection, and we still want to treat HOH as water because it is a typical PDB name. In the code where we do not have "water" in the selection, it will take your selection so it will check _resname HOH and name O2_ without hydrogens, and then the distC will be different than we had before (for WatFinder paper).
> 
> I still don't understand this. HOH should be recognized as water and we should be able to put "water" in the selection string to catch it.
> 
> I also don't get why we use O2 or OH2 and then include the whole water molecule. It still feels like something that we should control at the level of the input selection and not through the internal code because it could give users unexpected results.

You are right, and we could drop that one _if_, but then we will have slightly different results than what we had in the WatFinder paper comparison. The results would be different because _distC_ should be bigger now compared to the previous analysis. 

In the code, we select O2, and we check its location and whether some other water molecules are close. We are also including hydrogens (that is why _water_ and not the _selection_)[code below]. If we use the selection 'resname HOH', ProDy will include all three atoms (H,O, H) and compute the center of the mass, and that localization will be checked. It is fine as well, but it will give slightly different results than what we have in the WatFinder tutorial, and that may confuse people. 

sel = coords_all.select(str(**selection**)+' within '+str(distC)+' of center',
                                center=coords_all.getCoords()[ii])

jamesmkrieger:
But the default selection is factoring that in and any user selection can too

selection = kwargs.pop('selection', 'water and name OH2')

jamesmkrieger:
We should probably use element O or something like that to be more general as different software use different atom names

karolamik13:
> We should probably use element O or something like that to be more general as different software use different atom names

We might want to use '_name "O.*"_' 
I see that p.select('resname "A.*"') is possible although I never use it (http://www.bahargroup.org/prody/manual/reference/atomic/select.html?highlight=select#module-prody.atomic.select) so that will work for any oxygen name.

jamesmkrieger:
Sounds reasonable in case the element field isn’t set properly. We should check if it always starts with O but I think so

karolamik13:
I have changed the oxygen selection - it is working fine with our previous examples -> MD trajectory from NAMD and PDBs analysis of aurora kinase A (both examples from WatFinder paper).

jamesmkrieger:
Great

jamesmkrieger:
I’ll still want to go through the whole code of the function and check where the oxygen selection is used

jamesmkrieger (approved):
ok, I think it works now

PR #1982, opened 2024-10-30
hpb.so for python 3.11 and 3.12

karolamik13:
I created two hpb.so files for the newest Python versions.


jamesmkrieger:
Great! We need to update the tests accordingly

jamesmkrieger (approved):

PR #1992, opened 2024-11-15
replace PiStack angle_min_RB and angle_max_RB with PS versions

jamesmkrieger:
It's just for the user input and then the function uses angle_min and angle_max internally anyway


jamesmkrieger:
Actually this makes a difference for using these kwargs with calculating all interactions

karolamik13 (approved):

PR #1994, opened 2024-11-18
add chain bar to showCumulativeInteractionTypes energy

jamesmkrieger:
Previously, we didn't have any chain bars and the residue labels aren't that sensible for multi chain cases (here they have different numbers in the two chains so it's ok but usually that may not happen):
interactions_cum_energy_old_9atm_RL

Now, we have chain bars and good residue+chain labels
interactions_cum_energy_chains_9atm_RL


karolamik13:
Great. Thank you!

PR #1998, opened 2024-11-20
add energy once

jamesmkrieger:
Before, we could add the  same energy multiple times if the same pair of amino acids had different atoms interacting


jamesmkrieger:
Now we get several energies being much lower. 

Before, the minimum energy was about -10
matrix_energy_RH_relax_old

Now, the minimum is -6.65
matrix_energy_RH_relax_fixed

karolamik13 (approved):

PR #2022, opened 2024-12-27
anmd fix for pca input

jamesmkrieger:
PCA has variances as eigvals, so we need to take the reciprocal to get equivalent eigvals for scaling.


anupam-banerjee (approved):
Looks good to me.

jamesmkrieger:
Thanks

PR #2029, opened 2025-01-24
Parallel insty fix

jamesmkrieger:
I think this should work now using starmap instead


jamesmkrieger:
This didn't work at all

PR #2033, opened 2025-02-03
package mmtf test files and hpb.so with prody

jamesmkrieger:
We didn't notice this problem because we were using development installs in place but after making the conda-forge release, which installs fine, I noticed that I can't import the tests because the mmtf files are missing.


karolamik13 (dismissed):

AnthonyBogetti (dismissed):

jamesmkrieger:
@AnthonyBogetti, I guess this is working well in your new conda release attempts?

AnthonyBogetti:
@jamesmkrieger Let me test once more just to be sure. Then I will approve.

PR #2041, opened 2025-02-12
Replace np matrix

jamesmkrieger:
This will help with np matrix being deprecated.

We still need to check if there are other places


jamesmkrieger:
Maybe this is already done somewhere else but I'm not sure

Issue #1971, opened 2024-10-09
update clustenm for openmm 8


jamesmkrieger:
I spotted when trying to integrate ANMD that ClustENM is having problems with openmm 8.1.2, so we should stick to openmm 7.6 for now

The first problem is that we can't give platform properties is platform is none:
```
clustenm.run(n_modes=2, n_confs=5, n_gens=2, maxclust=2)
@> Kirchhoff was built in 0.03s.
@> Generation 0 ...
@> Minimization, heating-up & simulation in generation 0 ...
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[26], line 1
----> 1 clustenm.run(n_modes=2, n_confs=5, n_gens=2, maxclust=2)

File ~/software/scipion3/software/em/prody-github/ProDy/prody/dynamics/clustenm.py:1186, in ClustENM.run(self, cutoff, n_modes, gamma, n_confs, rmsd, n_gens, maxclust, threshold, solvent, sim, force_field, temp, t_steps_i, t_steps_g, outlier, mzscore, **kwargs)
   1184     LOGGER.info('Minimization in generation 0 ...')
   1185 LOGGER.timeit('_clustenm_min')
-> 1186 potential, conformer = self._min_sim(self._atoms.getCoords())
   1187 if np.isnan(potential):
   1188     raise ValueError('Initial structure could not be minimized. Try again and/or check your structure.')

File ~/software/scipion3/software/em/prody-github/ProDy/prody/dynamics/clustenm.py:320, in ClustENM._min_sim(self, coords)
    317 except ImportError:
    318     raise ImportError('Please install PDBFixer and OpenMM 7.6 in order to use ClustENM.')
--> 320 simulation = self._prep_sim(coords=coords)
    322 # automatic conversion into nanometer will be carried out.
    323 # simulation.context.setPositions(coords * angstrom)
    325 try:

File ~/software/scipion3/software/em/prody-github/ProDy/prody/dynamics/clustenm.py:303, in ClustENM._prep_sim(self, coords, external_forces)
    300 elif self._platform in ['CUDA', 'OpenCL']:
    301     properties = {'Precision': 'single'}
--> 303 simulation = Simulation(modeller.topology, system, integrator,
    304                         platform, properties)
    306 simulation.context.setPositions(modeller.positions)
    308 return simulation

[...]

PR #2096, opened 2025-03-11
ClustENM compatibility with OpenMM v8 and above

AnthonyBogetti:
This allows ClustENM to be used with versions of OpenMM over 8. The issue was a unit mismatch in the tolerance parameter for energy minimization. Something must have changed on their end going from v7 to v8.

### DIFF
--- a/prody/dynamics/clustenm.py
+++ b/prody/dynamics/clustenm.py
@@ -313,7 +313,7 @@ class ClustENM(Ensemble):
 
         try:
             from openmm.app import StateDataReporter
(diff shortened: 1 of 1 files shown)


AnthonyBogetti:
Fixes and completes #1971.

jamesmkrieger (approved):
It sounds reasonable to me. Hopefully this still works for OpenMM v7 too. Otherwise, we may need a try/except

AnthonyBogetti:
I tested it with OpenMM 7.7 and it still works. I will go ahead and merge it. Thank you for reviewing.

jamesmkrieger:
great. You're welcome

PR #2118, opened 2025-04-27
add rtb app

jamesmkrieger:
It seems much slower than using it from the API, but it does work.


jamesmkrieger:
> It seems much slower than using it from the API, but it does work.

I fixed this problem now! It was that the maximum block size could be the whole protein

AnthonyBogetti (approved):

AnthonyBogetti:
@jamesmkrieger If this is good, go ahead and merge.

PR #2120, opened 2025-04-27
try/except fromstring or frombuffer

jamesmkrieger:
I think something changes in some version of numpy that means that we need this


jamesmkrieger:
Maybe it's just to suppress deprecation warnings at this point. I'm not sure

PR #2126, opened 2025-05-08
Fix parseMMCIF and fetchPDB to get CIF files

jamesmkrieger:
Bug 1: parseMMCIF was taking .pdb files if they were found. This is solved by adding the format to the second run of findPDBFiles in fetchPDB. 

Bug 2: non-pdb files couldn't be retrieved from the PDBe as the extension replacement wasn't being done. This is fixed by making the replacement from .ent as well as .pdb


jamesmkrieger:
probably we need more tests for this

jamesmkrieger:
This won't fix the pfam gzip error, but it will at least make things slightly less cryptic

PR #2127, opened 2025-05-08
add wwpdb ftp and http

jamesmkrieger:
The wwPDB consortium itself also has these places for accessing pdb and mmcif files, besides the ones for the members that we already support. I found this by looking in the scipion code as they take files from there.


jamesmkrieger:
this needs tests too

PR #2128, opened 2025-05-08
Handle chimerax cif (missing fields)

jamesmkrieger:
There was a string of errors from trying to parse header information, such as biomols, from a chimerax cif file generated by scipion, which is now included in the tests datafiles. These commits should fix it.


AnthonyBogetti:
Hi @jamesmkrieger, go ahead and merge this one if ready and then I will review #2139.

PR #2137, opened 2025-06-10
Write mmcif header

jamesmkrieger:
Now we can add a full mmCIF header to a new set of coordinates, e.g.
```
In [1]: from prody import *

In [2]: ag1, header1 = parseMMCIF('D_1292146224_model-review_P1.cif', header=True)
@> 30156 atoms and 1 coordinate set(s) were parsed in 6.96s.

In [3]: ag2 = parsePDB('refined-coot-3.pdb')
@> 29673 atoms and 1 coordinate set(s) were parsed in 0.29s.

In [4]: writeMMCIF('D_1292146224_model-review_P1_chimera-coot-6-refmac-coot-3-prody.cif', ag2, header=header1)
```


jamesmkrieger:
Actually, this isn't quite there

jamesmkrieger:
This now more or less works but it is very very slow. It also doesn't have anything about the order of non-loop tables and the starDict that I've made is not a useful structure for interrogating the metadata. 

I'll probably improve it further to address these 3 points

karolamik13 (approved):
I checked the code before (old_*) and after (new_*) this improvement.

-rw-rw-r-- 1 xxx xxx 423K lip 29 20:36 old_5ir5_test_header.cif
-rw-rw-r-- 1 xxx xxx 556K lip 29 20:37 new_5ir5_test_header.cif 

The new one contains more information (header).

**The code:**
from prody import *

ag1, header1 = parseMMCIF('5IR5', header=True)
ag2 = ag1.copy()
writeMMCIF('5ir5_test_header.cif', ag2, header=header1)

karolamik13:
Screenshot from 2026-07-29 22-42-35

jamesmkrieger:
Great! And it’s not too slow?

karolamik13:
> Great! And it’s not too slow?

I noticed that it is noticeable slower. With one structure it is not a problem, but with hundrads that might be a problem.

PR #2140, opened 2025-06-28
py3dmol visualization of interactions

dkoes:
Simple implementation of interaction visualization.  Have not tested ligand interactions since the tutorial is broken (http://www.bahargroup.org/prody/tutorials/insty_tutorial/ligand.html?highlight=showligandinteraction_vmd).

### DIFF
--- a/prody/proteins/interactions.py
+++ b/prody/proteins/interactions.py
@@ -25,7 +25,7 @@ from prody.atomic import flags, sliceAtomicData
 from prody.utilities import importLA, checkCoords, showFigure, getCoords
 from prody.measure import calcDistance, calcAngle, calcCenter
 from prody.measure.contacts import findNeighbors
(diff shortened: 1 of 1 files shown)


jamesmkrieger:
> Hi @jamesmkrieger, is this ready to merge?

No idea. We'll have to check it and see.

karolamik13:
Yes. It was working fine. The only improvement I would do it selection of particular residues. Currently all residues are displayed as sticks too, but this is also fine. 

Tutorial might not work because plip was not installed which is required to obtain the results.

AnthonyBogetti:
Thank you, @karolamik13.

jamesmkrieger:
> Yes. It was working fine. The only improvement I would do it selection of particular residues. Currently all residues are displayed as sticks too, but this is also fine.

Ok, this is an improvement that can be made now that we have it merged. Thanks

PR #2150, opened 2025-08-06
Various fixes related to new dependency versions

jamesmkrieger:
- Replacement or fixing of numpy fromstring to not trigger binary mode that is removed in numpy 2.3
- Adding more try/except blocks for different attribute names for gap opening and extension in biopython 1.84 and later
- Removing numpy chararray as deprecated so soon to be removed

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,7 +33,7 @@ jobs:
           conda activate test
           conda install --yes -c conda-forge compilers gfortran
           conda install --yes \
-            "numpy<2.0" \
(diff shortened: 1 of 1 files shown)


jamesmkrieger:
> numpy still 1.26 here, I need to remove the numpy < 2.0 in the github workflow.

Ah ok

jamesmkrieger:
It’s now using numpy 2.2.6 for python 3.10 on linux and numpy 2.3.2 for python 3.12, so let’s see what happens

PR #2159, opened 2025-08-14
added pypi shield with links

AnthonyBogetti:
Adds back in pypi shield with version to README.rst since we now have a pypi install. Also added back in links to the shields.


jamesmkrieger (approved):
sounds good to me. Let's not merge it yet until we have the PyPI working

AnthonyBogetti:
I'll just add this in with #2160

PR #2165, opened 2025-08-19
a more generalized solution for cif parsing

anupam-banerjee:
Fix _getBiomoltrans to robustly parse mmCIF oper_expression: accept alphanumeric op IDs (X0, P), expand ranges anywhere (incl. inside lists), and handle products of groups like (X0)(1-5).

### DIFF
--- a/prody/proteins/cifheader.py
+++ b/prody/proteins/cifheader.py
@@ -182,69 +182,135 @@ def getCIFHeaderDict(stream, *keys, **kwargs):
 
 
 def _getBiomoltrans(lines):
+    from itertools import product
 
-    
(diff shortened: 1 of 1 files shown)


jamesmkrieger (approved):
looks good and I've tested pieces and it seems to work as expected. 

It would be good to have a unit test for this too

AnthonyBogetti (changes requested):
Let's come up with a unit test for this before merging.

AnthonyBogetti:
[test_ciffile.py](https://github.com/user-attachments/files/23913539/test_ciffile.py)
Here is a modified test_ciffile.py to cover the changes in this PR. @anupam-banerjee, please test on your code and if it works, let's merge this.

AnthonyBogetti (approved):

PR #2179, opened 2025-10-09
New hinge identification method

stoneandbone:
[HZ384] Update: Enhanced Hinge Identification Function

This update refines the hinge site detection algorithm based on the methodology of [HZ384] H. Zhang, M. Gur, I. Bahar (2024), PNAS 121(49): e2414333121.

**Summary of changes:**

1. Adaptive mode selection: If a GNM model is provided without specifying n_modes, the number of modes is now automatically determined to achieve 33% cumulative variance. When GNM modes are supplied, all provided modes are used for hinge detection.
2. Hinge detection: Hinges are identified from crossover points and residues within a defined threshold band in the eigenvector.
3. Region merging: Overlapping or adjacent hinge regions are merged, with optional user-defined spacing tolerance.
4. Transient reduction: Transient hinge regions are filtered out to improve robustness.
5. Terminal trimming: Hinges at N- or C-terminal ends, where loops may be presented, may be trimmed based on user preference.
6. Output: Returns a list of hinges by mode for downstream analysis.

### DIFF
--- a/prody/dynamics/analysis.py
+++ b/prody/dynamics/analysis.py
@@ -683,32 +683,32 @@ def getHinges(v, threshold=15, space=None):
                 j += 1
             regs.append(list(range(start, end + 1)))
     
-      ### Merge overlapping or adjacent regions (separated by space value) ###
-                
-        regs = sorted(regs, key=lambda x: x[0])
-        merged = [regs[0]]
-        s = 1 + space if space is not None else 0
-        
-        for curr in regs[1:]:
-            prev = merged[-1]
-            if curr[0] <= prev[-1] + s:
-                merged[-1] = list(range(prev[0], max(prev[-1], curr[-1]) + 1))
-            else:
-                merged.append(curr)
-
-        ### Select hinges from regions ###
-        fil=[]
-        for reg in merged:
-            av = np.abs(v[reg])
+    ### Merge overlapping or adjacent regions (separated by space value) ###
             
-            if len(reg) >= s + 4: # region with length of 2+ discrete regions
-                fil.extend([reg[i] for i in np.argsort(av)[:2] if av < band]) # end hinges within band
-            else:
-                if v[reg[0]] * v[reg[-1]] > 0 : # skip transient region
-                    continue
-                fil.append(reg[np.argmin(av)]) # minimum point as hinge
+    regs = sorted(regs, key=lambda x: x[0])
+    merged = [regs[0]]
+    s = 1 + space if space is not None else 0
+    
+    for curr in regs[1:]:
+        prev = merged[-1]
+        if curr[0] <= prev[-1] + s:
+            merged[-1] = list(range(prev[0], max(prev[-1], curr[-1]) + 1))
+        else:
+            merged.append(curr)
 
-        return [int(x) for x in fil]
+    ### Select hinges from regions ###
+    fil=[]
+    for reg in merged:
+        av = np.abs(v[reg])
+        
+        if len(reg) >= s + 4: # region with length of 2+ discrete regions
+            fil.extend([reg[i] for i in np.argsort(av)[:2]]) # end hinges within band
+        else:
+            if v[reg[0]] * v[reg[-1]] > 0 : # skip transient region
+                continue
+            fil.append(reg[np.argmin(av)]) # minimum point as hinge
+
+    return [int(x) for x in fil]
     
 
 def _getModeVecs(gnm, n_modes=None, min_variance=0.33, atoms=None):
@@ -794,7 +794,7 @@ def getGlobalHinges(gnm, n_modes=None, threshold=15, space=None, atoms=None, min
         fst = 0
         for chain in hv.iterChains():
             n = chain.numAtoms()
-            chains.append((fst, fst + n))
+            chains.append((fst, fst + n - 1))
             fst += n
     else:
         chains = [(0, p)]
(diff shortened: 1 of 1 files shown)


AnthonyBogetti:
The (random) test failures here are due to the pfam tests which should be fixed in #2189

jamesmkrieger:
> The (random) test failures here are due to the pfam tests which should be fixed in #2189 maybe we can merge that one first.

I agree

prody/dynamics/analysis.py
```diff
@@ -646,6 +646,143 @@ def identify(v):
...
+        1) If GNM model is provided without specification of n_modes, number of modes will be determined to achieve 33% cumulative variance.
+           If selected GNM modes are provided directly, all modes will be used for hinge detection.
+        2) Identify hinge regions based on crossovers and residues within band, whose magnitude is specified by the threshold.
+        3) Merge overlapping or adjacent hinge regions.
+        4) Reduce transient hinge regions.
+        5) Trim hinges at N- or C-terminal ends. 
+        6) Return list of hinges by mode.
+    
+    :param modes: GNM model or ModeSet
+    :type modes: GNM or Mode instance
```
jamesmkrieger: Use the format that sphinx can read like in other doc strings. Also put ModeSet again in type and put a more meaningful description than the types in the bit above, linking to what was said above about selecting modes. 

Double check that param is ok too. We usually put arg

prody/dynamics/analysis.py
```diff
@@ -646,6 +646,143 @@ def identify(v):
...
+    :param n_modes: Number of modes to consider. Defaults to all.
+    :type n_modes: int or None
```
jamesmkrieger: Check if or is ok. We usually put a comma. Again remember this should be machine readable as well as human readable

jamesmkrieger: Maybe handle default of None in the bit above as well or instead e.g. defaults to all modes if left as **None**

prody/dynamics/analysis.py
```diff
@@ -646,6 +646,143 @@ def identify(v):
...
+        """Detect hinge-like regions within single eigenvector."""
+        
+        crx = np.nonzero(np.diff(np.sign(v)))[0] 
+        band = np.sqrt(1 / len(v)) / threshold
+        regs = []
+
+        for i in crx:
+            r = [i, i + 1]
+   
+            if abs(v[i]) > band and abs(v[i + 1]) > band:
```
jamesmkrieger: Again simplify nested if else blocks. ChatGPT suggests this. Please check if it makes sense 
```
band = np.sqrt(1 / len(v)) / threshold
regs = []

for i in crx:
    # Start hinge region around the sign change
    start = i
    end = i + 1

    # Expand left
    j = start
    while j >= 0 and abs(v[j]) < band:
        start = j
        j -= 1

    # Expand right
    j = end
    while j < len(v) and abs(v[j]) < band:
        end = j
        j += 1

    # Store inclusive region
    regs.append(list(range(start, end + 1)))
```

prody/dynamics/analysis.py
```diff
@@ -646,6 +646,143 @@ def identify(v):
...
+    :param n_modes: Number of modes to consider. Defaults to all.
+    :type n_modes: int or None
+    
+    :param threshold: Threshold controlling band width for hinge region identification.
+    :type threshold: float
+    
+    :param space: spacing between hinges. Defaults to None. Higher space value will minimize the number of local hinges while retaining global hinges.
+    :type space: int or None
+    
+    :param trim: Whether to hinges at the N- or C-terminal ends by Protein length // 20. Defaults to False.
```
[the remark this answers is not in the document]
jamesmkrieger: Having atoms as inputs would also allow you to handle multiple chains, which is currently lacking. At the moment there would be merging across chain boundaries

prody/dynamics/analysis.py
```diff
@@ -646,6 +646,143 @@ def identify(v):
         return sorted(set(hinge_list))
     return hinges
 
+
+
+def hingeFinder(gnm, n_modes=None, threshold=15, space=None, trim=False):
+    """
+    [HZ384] H Zhang, M Gur, I Bahar (2024) Global hinge sites of proteins as target sites for drug binding Proc Natl Acad Sci USA 121 (49), e2414333121 
+    Updated hinge identification based on [HZ384]. This will:
+        1) If GNM model is provided without specification of n_modes, number of modes will be determined to achieve 33% cumulative variance.
```
jamesmkrieger: I guess a different percentage could also be used as a decimal input to the number of modes

prody/dynamics/analysis.py
jamesmkrieger: A user could also provide vectors as inputs directly

AnthonyBogetti:
Miraculously, the current tests all passed.

prody/dynamics/analysis.py
```diff
@@ -748,36 +698,121 @@ def detectHinges(v, threshold):
...
+    """
+    Return eigenvector array for hinge detection.
+
+    Handles GNM and ModeSet inputs, including reuse of precomputed modes.
+    """
+    if isinstance(gnm, ModeSet):
+        return gnm.getArray()
+
+    if not isinstance(gnm, GNMBase):
+        raise TypeError("Input must be a GNM model or ModeSet instance.")
```
jamesmkrieger: Maybe allow gnm=None if atoms is not None as a new gnm can be instantiated? People are unlikely to input a newly instantiated gnm object themselves

PR #2237, opened 2026-06-05
Prody CaviTracer - start point

karolamik13:
I added a start point for the detection of channels, tunnels, and pores within protein structure. Before, it was done in an automatic way. Now, we can also provide [x,y,z] as a starting point (for example, localization of the catalytic site) for the algorithm that will search the interior cavities.

Tests:
(1) Single PDB:
from prody import *
p = parsePDB('1tqn').select('protein')
calcChannels(p, output_path='test1.pdb', start_point=[-22.312, -20.065, -11.144])

(2) Multi-PDB:
from prody import *
p = parsePDB('multi.pdb').select('protein')
calcChannelsMultipleFrames(p, output_path='mm_test_1', start_point=[-2.353, -0.103, 5.668])

(3) DCD file:
from prody import *
PDBfile = 'structure.pdb'
DCDfile = 'trajectory.dcd'
atoms = parsePDB(PDBfile)
dcd = Trajectory(DCDfile)
dcd.link(atoms)
dcd.setCoords(atoms)
calcChannelsMultipleFrames(atoms, dcd, output_path='mm_test_2', start_point=[-2.353, -0.103, 5.668])

### DIFF
--- a/prody/proteins/channels.py
+++ b/prody/proteins/channels.py
@@ -562,7 +562,7 @@ def calcChannels(atoms, output_path=None, separate=False, start_point=None, r1=3
     return channels, [coords, s_srf.simp, merged_cavities, s_clr.simp]
 
             
-def calcChannelsMultipleFrames(atoms, trajectory=None, output_path=None, separate=False, **kwargs):
+def calcChannelsMultipleFrames(atoms, trajectory=None, output_path=None, separate=False, start_point=None, **kwargs):
     """Compute channels for each frame in a given trajectory or multi-model PDB file.
 
     This function calculates the channels for each frame in a trajectory or for each model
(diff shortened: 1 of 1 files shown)


prody/proteins/channels.py
```diff
@@ -401,6 +401,11 @@ def calcChannels(atoms, output_path=None, separate=False, r1=3, r2=1.25, min_dep
         are saved in a single PDB file. Default is False.
     :type separate: bool
 
+    :param start_point: Optional starting point for channel search. If provided, the algorithm will use 
+        the tetrahedron whose Voronoi vertex is closest to this point as the starting tetrahedron (overriding 
+        the default automatic seed selection based on the deepest tetrahedron). Coordinates must be given in Å.
+    :type start_point: array-like of shape (3,) or None
```
jamesmkrieger: check that this type is machine-readable

prody/proteins/channels.py
```diff
@@ -466,6 +473,16 @@ def calcChannels(atoms, output_path=None, separate=False, r1=3, r2=1.25, min_dep
     else:
         from pathlib2 import Path
     
+    if start_point is not None:
+        if not isinstance(start_point, list) or len(start_point) != 3:
```
jamesmkrieger: prody has a function for a more general array-like type that could be used here. Otherwise update the type in the docs to be a list

prody/proteins/channels.py
```diff
@@ -466,6 +473,16 @@ def calcChannels(atoms, output_path=None, separate=False, r1=3, r2=1.25, min_dep
...
+            raise ValueError(
```
jamesmkrieger, line 480: there should be separate errors for type and value

prody/proteins/channels.py
```diff
@@ -1588,7 +1617,21 @@ def differential_volume(t):
         
         return total_volume
             
-    
+    def set_starting_tetrahedra_from_point(self, cavities, vertices, start_point):
+        # start_point: array-like (3,)
```
jamesmkrieger: make this a doc string not a comment and correct for actual type as above

jamesmkrieger (approved):
looks good. Just a few comments on type docs and handling

karolamik13:
Thank you, James, for checking. I have incorporated your suggestions.

PR #2239, opened 2026-06-17
(by claude) Fix PDB sequence clusters: migrate off deprecated RCSB bc-*.out files

### LINKED ISSUE #2238: Bug report: `fetchPDBClusters` broken — RCSB deprecated the legacy `bc-*.out` cluster files
# Bug report: `fetchPDBClusters` broken — RCSB deprecated the legacy `bc-*.out` cluster files

## Description of the bug.
Below, please describe the error you encountered and what you were trying to do with ProDy.

I was trying to fetch PDB sequence clusters with `fetchPDBClusters()` and then use
`listPDBCluster()` to find the cluster for a chain.

`fetchPDBClusters` downloads from the legacy URL

```
https://cdn.rcsb.org/resources/sequence/clusters/bc-{sqid}.out
```

RCSB has **deprecated and removed these legacy `bc-*.out` files**. The URL now returns
an **HTTP 404** (an Apache "Not Found" HTML page), so no usable cluster data is ever
retrieved. Depending on the environment this surfaces in one of two ways:

1. **Download silently fails, then a later call crashes.** `openURL` raises on the 404,
   so `fetchPDBClusters` only logs the warning `"Clusters at N% sequence identity level
   could not be downloaded."` and writes nothing. A subsequent `loadPDBClusters` /
   `listPDBCluster` then crashes with `FileNotFoundError`, because after the failed
   re-fetch `loadPDBClusters` calls `os.path.getmtime(filename)` on a file that does not
   exist (`prody/proteins/pdbclusters.py:42-47` — it checks `os.path.isfile` *before*
   the fetch but never re-checks after).

2. **"Reports OK" but the saved file is unusable.** In environments where the request
   does not raise on the 404 (e.g. behind a proxy that returns a 200 with an error page),
   `fetchPDBClusters` writes the HTML error page into `bc-{sqid}.out.gz`, increments its
   success counter, and logs `"All selected PDB clusters were downloaded successfully."`
   The file is then a corrupt gzip / non-cluster data and fails when `loadPDBClusters`
   tries to parse it.

The new data is served at:

```
https://cdn.rcsb.org/resources/sequence/clusters/clusters-by-entity-{sqid}.txt   (HTTP 200, text/plain)
```

Note this also changes the **record format**: the legacy `bc` files used
`PDBID_CHAIN` (e.g. `1XXX_A`), whereas the new `clusters-by-entity` files use
`PDBID_ENTITY` (e.g. `1XXX_1`). So `listPDBCluster`, which currently splits on `_` and
matches a `(pdb, chain)` tuple, will also need updating to map entity IDs to chains once
the download URL is fixed.

## Do you have any error messages or logs?
Below, please post the error you got and text from any relevant log files.

Fetch + use (clean environment, download raises on 404):

```
@> WARNING Clusters at 40% sequence identity level could not be downloaded.
@> WARNING PDB clusters could not be downloaded.
@> Loading PDB sequence clusters for sequence identity 40.
@> WARNING Clusters at 40% sequence identity level could not be downloaded.
@> WARNING PDB clusters could not be downloaded.
Traceback (most recent call last):
  File "<string>", line 7, in <module>
  File ".../prody/proteins/pdbclusters.py", line 97, in listPDBCluster
    loadPDBClusters(sqid)
  File ".../prody/proteins/pdbclusters.py", line 47, in loadPDBClusters
    diff = (time.time() - os.path.getmtime(filename)) / 604800.
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen genericpath>", line 67, in getmtime
FileNotFoundError: [Errno 2] No such file or directory: '/home/USER/.prody/pdbclusters/bc-40.out.gz'
```

Direct confirmation that the legacy URL is gone and the new one works:

```
$ curl -sS -o /dev/null -w "%{http_code} %{content_type}\n" \
    https://cdn.rcsb.org/resources/sequence/clusters/bc-40.out
404 text/html; charset=iso-8859-1

$ curl -sS -o /dev/null -w "%{http_code} %{content_type} %{size_download}\n" \
    https://cdn.rcsb.org/resources/sequence/clusters/clusters-by-entity-40.txt
200 text/plain 22429241
```

First bytes of the new file (note `PDBID_ENTITY` format):

```
10EX_4 10FF_4 10OP_5 10OP_6 10OR_5 10OR_6 10ZO_1 11HA_3 ...
```

## What is your setup?
Below, please describe your setup. What operating system are you using (MacOS, Linux, etc.).

- OS: Rocky Linux 8.7 (Green Obsidian), Linux x86_64 (glibc 2.28)
- Python: 3.12.12

## How did you install ProDy?
Below, please say whether you installed from source, with conda or with pip. Please also add what version of ProDy you are using.

- Installed with pip (into a uv-managed virtual environment)
- ProDy version: **2.6.1**

## What did your code look like?
Below, if possible, please post the exact code you ran that led to the error you received. Please also attach any relevant input files. This will help us more quickly respond you your issue.

```python
import prody
prody.confProDy(verbosity='info')

prody.fetchPDBClusters(40)
cluster = prody.listPDBCluster('1xxx', 'A', 40)
print(cluster)
```

## Anything else?
Below, please write any other details you would like us to know about.

Suggested fix, in `prody/proteins/pdbclusters.py`:

1. Point `fetchPDBClusters` at the current endpoint
   `https://cdn.rcsb.org/resources/sequence/clusters/clusters-by-entity-{sqid}.txt`
   instead of `bc-{sqid}.out`.
2. Validate the HTTP response (status and/or content type) before writing, so an HTML
   error page is never silently saved as `*.gz` and reported as a successful download.
3. In `loadPDBClusters`, re-check `os.path.isfile(filename)` after calling
   `fetchPDBClusters(sqid)` and raise a clear error (instead of crashing in
   `os.path.getmtime`) when the download did not succeed.
4. Update `listPDBCluster`/parsing to account for the new `PDBID_ENTITY` record format
   (vs. the old `PDBID_CHAIN`).

jamesmkrieger:
Fixes #2238 (discovered and written by claude too)

RCSB removed the legacy bc-{sqid}.out cluster files (now HTTP 404), so fetchPDBClusters retrieved no usable data and a subsequent loadPDBClusters crashed in os.path.getmtime with FileNotFoundError.

- fetchPDBClusters: download from the current clusters-by-entity-{sqid}.txt endpoint, and refuse to save empty bodies or HTML error pages (so a failed download is never reported as success).
- loadPDBClusters: re-check the cache file after fetching and raise a clear IOError instead of crashing in getmtime; parse on the last underscore so identifiers containing underscores (computed structure models, e.g. AF_AFP12345F1) still yield clean (identifier, entity) pairs.
- listPDBCluster: the data now clusters by polymer entity rather than chain, so the second argument is an entity id; update matching and docstrings.
- Add tests for fetch/list and the failed-download error path.


karolamik13 (approved):
I have tested it. Everything is working fine.

PR #2241, opened 2026-06-26
CaviTracer - Optimization

MatthewLicht:
I modified the sphere_fit, delete_simplices3d ,delete_section, surface_layer, and dijkstra functions in the ChannelCalculator Class for faster processing. Additionally, calcChannelsMultipleFrames uses multiprocessing Pool for parallel processing. 

to test the code
from prody import *

Tests:
(1) Single PDB:
from prody import *
p = parsePDB('1tqn').select('protein')
calcChannels(p, output_path='test1.pdb')

(2) Multi-PDB:
from prody import *
p = parsePDB('multi.pdb').select('protein')  (e.g. '2L6X')
calcChannelsMultipleFrames(p, output_path='mm_test_1', max_proc=2)

(3) DCD file:
from prody import *
PDBfile = 'structure.pdb'
DCDfile = 'trajectory.dcd'
atoms = parsePDB(PDBfile)
dcd = Trajectory(DCDfile)
dcd.link(atoms)
dcd.setCoords(atoms)
calcChannelsMultipleFrames(atoms, dcd, output_path='mm_test_2',max_proc=2)

(4) Multiple AtomGroups:
from prody import *
pdbs = [list of pdb files]
ags = [parsePDB(pdb) for pdb in pdbs ]
calcChannelsMultipleAtomGroups(ags)

PR #2249, opened 2026-07-16
CaviTracer - refactor for enhanced accuracy and speed of channel identification to be competitive with established SOTA tools

briza81:
**Accuracy/correctness improvements in the current branch over the original implementation and compared to CAVER3**
## 1. Voronoi diagrams (VD) - the atoms are no longer treated as equal-radius points

The original ran a plain Delaunay/Voronoi tessellation on atom centers (and auto-excluded hetero/H atoms). Now there's a diagram argument with three modes:
- "homogenized" (new default) - each atom is replaced by concentric shells of equal-radius balls (max_deviation controls fidelity), so an ordinary tessellation approximates the additively-weighted (Apollonius) diagram within the max_deviation tolerance (MolAxis/CAVER3 trick)
- "simple" - the original point-based behavior.
- "weighted" - the true Apollonius diagram via the new _vorpy_aw.py (third-party vorpy + optional numba kernel, spatial box decomposition, content-keyed on-disk .npz cache, weighted_cache=True). However, this is super slow, circa 100x over high-accuracy homogenized VD
-> Because atom radii are now properly treated, we no longer need to drop hetero/H atoms.

## 2. Width/bottleneck/Cost function- now they are sampled at the gates, not just the vertices
The original measured channel clearance only at Voronoi vertices (circumcenters), which are local maxima of width. Here we move to a per-edge gate clearance (min over the shared Delaunay face, closed-form point-to-segment) + integrating the clearance along the actual edge:
- Dijkstra cost is now driven by the gate clearance, making the routing cost direction-symmetric.  Edge clearance integral provides the ultimate accuracy for the edge cost. 
- Reported bottleneck now takes the min clearance on the edge.
- Volume tube injects a gate knot midway between vertices so the radius profile dips where the channel is actually narrow.

## 3. limiting mesh sensitivity - avoiding layer-based metrices that change with VD density, which changes with max_deviation
- migrate to physical metrics - distance/topological distance, instead of relying on layers, which are r1, r2 and max_deviation dependent 
- hence, depth is now measured in geodesic Ångström, not tetrahedron layers -> a multi-source Dijkstra along Voronoi edges (with a scale-invariant). Defaults recalibrated to match old behavior for both channels and surface cavities (min_depth 10→5 Å, etc.).

## 4. surface/cavity definition - local enclosure-based peel/erosion
- The original erosion depended on the tessellation and on r1. Now, a boundary tetrahedron is stripped only while it's open (based on ray-marched directions that hit protein within range), so burial is measured against real atoms. r1 only decides where erosion starts. 
- Importantly, mouths absorb instead of conduct. Exit tetrahedra are no longer pre-sampled targets (thinned by sparsity); before the Dijkstra search, every mouth's outgoing edges are deleted so a surface tetrahedron, a probe of radius r2 can leave through becomes a sink (incoming edges are kept, outgoing edges are removed). This makes "a channel ends at its first surface contact" a hard constraint of the search rather than a cut applied afterward, avoiding the mesh-dependent failure where a genuine narrow interior corridor lost the argmin to a wide surface groove and vanished from the output. Now, every reachable mouth emits a candidate, and their deduplication decides which mouths are one opening: when two channels reach the same opening, we test if they have similar corridor/path to this opening (similarity within route_tolerance), and if so => two channels reaching the same exit via the same path -> we keep only the cheaper one. ->
- Therefore, sparsity is not needed anymore to control against channel explosion. It is now an argument to fine-tune the minimal required distance between channel openings of distinct channels.

## 5. start-point handling
- restrict_channels_to_start_point limits the search only to the anchored cavity
- start_point_search - optimize the seed to the widest nearby tetrahedron of at least the same depth, which is accessible from the start point, instead of using the nearest one

## 6. performance optimization
Here we have several overlaps with https://github.com/prody/ProDy/pull/2241 by Matthew, to name a few:
vectorized tessellation filtering, surface erosion restricted to the boundary shell (O(n)→O(n^⅔) per pass), circumcenters recovered from the Delaunay paraboloid lifting (skips a second Qhull pass), one CSR weighted graph + a single scipy multi-target Dijkstra per cavity (replacing per-(seed, exit) heap Dijkstra), vectorized Simpson volume, and ~30× faster vectorized buildSparseGraph

## 7. output & reporting
- channels now carry a per-channel cost (Dijkstra path weight) and curvature, are ordered by ascending cost (channel 0 = best tunnel), and each gets a REMARK line in the PQR
- multi-channel PQR now has corrected CONNECT items, and separate resIDs - same as the PQR files for cavities had. 
- added per-stage LOGGER timing
- local _warn() workaround for a global filter that was swallowing WARNINGs

## 8. Recalibrated defaults to losely agree with CAVER3 outputs
r2 1.25→0.9, min_depth 10→5, bottleneck 1→0.0, sparsity 15→1, plus the new diagram/max_deviation/min_enclosure args

## **More details, execution time measurements, and use cases for accuracy are available in the attached file: [Refoactored_CaviTracer_comparisons.pdf](https://github.com/user-attachments/files/30074248/Refoactored_CaviTracer_comparisons.pdf)**

### DIFF
--- a/prody/proteins/channels.py
+++ b/prody/proteins/channels.py
@@ -711,7 +711,8 @@ def showSurfaceCavities(surface, cavities=None, model=None, show_surface=False,
     o3d.visualization.draw_geometries(meshes_to_visualize)
 
 def calcChannels(atoms, output_path=None, separate=False, start_point=None,
-    restrict_channels_to_start_point=True, r1=3, r2=0.9, min_depth=10, 
+    restrict_channels_to_start_point=True, start_point_search=3.0,
+    r1=3, r2=0.9, min_depth=10,
     min_volume=None, max_volume=None, max_depth=None, bottleneck=0.0, 
     sparsity=1, min_tetrahedra=None, max_tetrahedra=None, cavities_only=False, 
     diagram="homogenized", max_deviation=0.1, truncate_at_surface=True,
@@ -747,24 +748,37 @@ def calcChannels(atoms, output_path=None, separate=False, start_point=None,
          False.
     :type separate: bool
 
-    :arg start_point: Optional starting point for channel search. This can be 
-        either a 3D coordinate point or an atomic selection/AtomGroup. If the 
-        3D coordinate point will be provided, the algorithm will use the 
-        tetrahedron whose Voronoi vertex is closest to this point as the 
-        starting tetrahedron (overriding the default automatic seed selection 
-        based on the deepest tetrahedron). Coordinates must be given in Å.
+    :arg start_point: Optional starting point for channel search. This can be
+        either a 3D coordinate point or an atomic selection/AtomGroup. If the
+        3D coordinate point will be provided, the algorithm seeds the channel
+        search near this point (overriding the default automatic seed selection
+        based on the deepest tetrahedron); see ``start_point_search`` for how
+        the seed tetrahedron itself is picked. Coordinates must be given in Å.
         If an atomic selection is provided, its geometric center is used as the
          starting point.
     :type start_point: list, tuple, or ndarray (length 3), :class:`.Atomic`, or None
 
-    :arg restrict_channels_to_start_point: Only used when ``start_point`` is 
-        provided. If True (default), the channel search is restricted to the 
-        single cavity whose closest tetrahedron is globally nearest to 
-        ``start_point``, so  channels are computed only for the region around 
-        that point instead of one channel bundle per detected cavity. If False, 
-        ``start_point`` merely overrides the seed (starting) tetrahedron of 
+    :arg restrict_channels_to_start_point: Only used when ``start_point`` is
+        provided. If True (default), the channel search is restricted to the
+        single cavity whose closest tetrahedron is globally nearest to
+        ``start_point``, so  channels are computed only for the region around
+        that point instead of one channel bundle per detected cavity. If False,
+        ``start_point`` merely overrides the seed (starting) tetrahedron of
         every cavity and channels are still computed for all cavities.
-    :type restrict_channels_to_start_point: bool 
+    :type restrict_channels_to_start_point: bool
+
+    :arg start_point_search: Only used when ``start_point`` is provided. Radius,
+        in Angstrom, of the neighbourhood of ``start_point`` searched for the seed
+        tetrahedron. The tetrahedron nearest ``start_point`` is often a tight one,
+        and since every channel of the cavity starts there, its inscribed radius
+        caps all of their bottlenecks and appears as one shared bottleneck at the
+        joint beginning of the bundle. Seeded instead is the widest tetrahedron within
+        ``start_point_search`` of ``start_point`` that belongs to the same cavity, is
+        no shallower than the nearest one (so the seed cannot drift out towards the
+        mouth) and is reachable from it through that neighbourhood (so it stays in the
+        void the point sits in rather than crossing a wall). Default is 3.0; use 0 to
+        seed the nearest tetrahedron as-is.
+    :type start_point_search: float
 
     :arg r1: The first radius threshold used during the deletion of simplices, 
         which is used to define the outer surface of the channels. Default is 3
@@ -1207,7 +1221,8 @@ def calcChannels(atoms, output_path=None, separate=False, start_point=None,
     if start_point is not None:
         c_surface_cavities = calculator.setStartingTetrahedraFromPoint(
             c_surface_cavities, s_clr.verti, start_point, coords, vdw_radii,
-            s_clr.simp, restrict_channels_to_start_point)
+            s_clr.simp, s_clr.neigh, restrict_channels_to_start_point,
+            start_point_search)
 
     c_filtered_cavities = calculator.filterCavities(c_surface_cavities, min_depth)
     LOGGER.report('{0} surface cavities detected and filtered in %.2fs.'.format(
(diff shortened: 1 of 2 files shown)


briza81:
@karolamik13 and @MatthewLicht pls have a look, more details over email.

@jamesmkrieger Karolina suggested asking if you could review this, too thank you

prody/proteins/channels.py
```diff
@@ -51,43 +101,53 @@ def checkAndImport(package_name):
...
-    """Generates a 3D model of molecular structures using VMD and returns it as an Open3D TriangleMesh.
-
-    This function creates a temporary PDB file from the provided atomic data and uses VMD (Visual Molecular Dynamics)
-    to render this data into an STL file, which is then loaded into Open3D as a TriangleMesh. The function handles
-    the creation and cleanup of temporary files and manages the subprocess call to VMD.
+    """Generates a 3D model of molecular structures using VMD and returns it as
+      an Open3D TriangleMesh.
+
+    This function creates a temporary PDB file from the provided atomic data a
+    nd uses VMD (Visual Molecular Dynamics) to render this data into an STL
```
jamesmkrieger: Don’t split doc string mid word

jamesmkrieger:
Thank @briza81, from the summary these sound like great changes and I will take a look at the details soon

prody/proteins/channels.py
```diff
@@ -198,33 +262,41 @@ def getVmdModel(vmd_path, atoms, representation='NewCartoon'):
...
+    their spline representations. It can also display a molecular model (e.g., 
+    the protein structure) and a surface (e.g., cavity surface) in the same 
+    visualization. The function utilizes the Open3D library to create and 
+    render the 3D meshes.
 
     To install Open3D use: 
-    conda install open3d (for Anaconda users; version open3d-0.19.0 was used during the developement)
-    or pip install open3d
+    conda install open3d (for Anaconda users; version open3d-0.19.0 was used 
+    during the developement) or pip install open3d
```
jamesmkrieger: typo to fix here too as above

prody/proteins/channels.py
```diff
@@ -243,7 +315,8 @@ def create_mesh_from_spline(centerline_spline, radius_spline, n=5):
         centers = centerline_spline(t)
         radii = radius_spline(t)
 
-        spheres = [o3d.geometry.TriangleMesh.create_sphere(radius=r, resolution=20).translate(c) for r, c in zip(radii, centers)]
+        spheres = [o3d.geometry.TriangleMesh.create_sphere(radius=r,
```
jamesmkrieger: this one can be wrapped more cleanly

prody/proteins/channels.py
```diff
@@ -309,26 +383,31 @@ def create_mesh_from_spline(centerline_spline, radius_spline, n=5):
...
+    molecular structure.
+    It uses the Open3D library to render the cavities as a triangle mesh. 
+    Optionally, it can also display the molecular surface as a wireframe 
+    overlay.
 
     To install Open3D use: 
-    conda install open3d (for Anaconda users; version open3d-0.19.0 was used during the developement)
-    or pip install open3d
+    conda install open3d (for Anaconda users; version open3d-0.19.0 was used 
+    during the developement) or pip install open3d
```
jamesmkrieger: generally replace developement with development

prody/proteins/channels.py
```diff
@@ -626,98 +710,330 @@ def showSurfaceCavities(surface, cavities=None, model=None, show_surface=False,
...
+        MolAxis and CAVER 3
+        "simple" - the original atoms are used with their individual van der 
+        Waals radii directly. This is very inaccurate and should be avoided in 
+        almost all cases.
+        "weighted" - the true additively-weighted (Apollonius) Voronoi diagram is
+        built directly from the atoms with their individual van der Waals radii,
+        using the third-party ``vorpy`` package (with a compiled kernel when
+        ``numba`` is available). This is the exact diagram the "homogenized" mode
+        approximates, at a higher cost (~ 100x slower, for 4000 atoms).
+
```
jamesmkrieger: no line space here

prody/proteins/channels.py
```diff
@@ -763,67 +1077,222 @@ def calcChannels(atoms, output_path=None, separate=False, start_point=None, r1=3
...
+                         "default for diagram='weighted') or None.")
+
+    calculator = ChannelCalculator(atoms, r2=r2, sparsity=sparsity,
+                                   route_tolerance=route_tolerance,
+                                   edge_cost=edge_cost)
+
+    elements = np.char.upper(np.asarray(atoms.getElements(), dtype=str))
+    has_hydrogens = bool(np.any(elements == 'H'))
+
+    # An X-ray structure carries no hydrogens, and its carbons keep their full vdW
```
jamesmkrieger: does this generalise to cryo-EM too? Also write "generally carries no hydrogens". I think even some crystal structures at very high resolution eventually have some.

prody/proteins/channels.py
```diff
@@ -2275,98 +2852,337 @@ def __init__(self, tetrahedra, is_connected_to_surface):
...
+        self.edge_cost = edge_cost
+        # Filled once by buildSparseGraph and read by the channel geometry:
+        # the per-simplex Voronoi-vertex clearance (the spline knots) and the
+        # per-edge gate clearance on each shared Delaunay face (the reported
+        # bottleneck and, later, the Dijkstra cost). Cached so the two consumers
+        # share one definition of width instead of recomputing it apart.
+        self._vertex_clearance = None
+        self._edge_bottleneck = None
+
+    # def sphereFit(self, vertices, tetrahedron, vertice, vdw_radii, r):
```
jamesmkrieger: Best to uncomment or delete commented code. I can't really see a good reason for keeping code if you don't want people to be able run it unless it's needed for the record but doesn't actually run. If there are experiments on it that will be mentioned in the paper, then it's better to uncomment it for easier reproducibility

prody/proteins/channels.py
```diff
@@ -2275,98 +2852,337 @@ def __init__(self, tetrahedra, is_connected_to_surface):
...
         
-    def sphere_fit(self, vertices, tetrahedron, vertice, vdw_radii, r):
-        center = vertice
-        d_sum = sum(np.linalg.norm(center - vertices[atom]) for atom in tetrahedron)
-        r_sum = sum(r + vdw_radii[atom] for atom in tetrahedron)
-        
-        return d_sum >= r_sum
+    #     return d_sum >= r_sum
+
+    def deleteSimplices3d(self, points, simplices, neighbors, vertices,
```
jamesmkrieger: This function needs a doc string

prody/proteins/channels.py
```diff
@@ -2275,98 +2852,337 @@ def __init__(self, tetrahedra, is_connected_to_surface):
...
+                          vdw_radii, r, surface):
+        simplices = np.asarray(simplices)
+        neighbors = np.asarray(neighbors)
+        vertices = np.asarray(vertices)
+
+        n = len(simplices)
+        if n == 0:
+            return simplices, neighbors, vertices
+
+        # Vectorized sphereFit: for each tetrahedron compare the sum of distances
```
jamesmkrieger: This could go in a helper function for easier validation of it's own results and reusability too, like its ancestor that is deprecated and currently commented

jamesmkrieger (approved):
Generally looks good. There are a few minor things of a more stylistic nature that I don't need to review again. I trust you know what you're doing on the details and it generally looks clear and well commented and documented

prody/proteins/channels.py
briza81: resolved in b96635ab1170fe0ef2e128c3728c4b5dc153a82d

prody/proteins/channels.py
briza81: resolved in b96635ab1170fe0ef2e128c3728c4b5dc153a82d

prody/proteins/channels.py
briza81: resolved in b96635ab1170fe0ef2e128c3728c4b5dc153a82d

prody/proteins/channels.py
briza81: resolved in b96635ab1170fe0ef2e128c3728c4b5dc153a82d

prody/proteins/channels.py
briza81: resolved in b96635ab1170fe0ef2e128c3728c4b5dc153a82d

prody/proteins/channels.py
briza81: resolved in 6862b7b3b166c17118b6870bf16a183f3b648bd0

prody/proteins/channels.py
briza81: resolved in 6862b7b3b166c17118b6870bf16a183f3b648bd0

prody/proteins/channels.py
briza81: resolved in 6862b7b3b166c17118b6870bf16a183f3b648bd0

prody/proteins/channels.py
briza81: resolved in b96635ab1170fe0ef2e128c3728c4b5dc153a82d

briza81:
@jamesmkrieger thank you for feedback, all should be resolved now
@karolamik13 should be ready for you

karolamik13:
After improvements from Jan and Matthew, calcChannelSurfaceOverlaps() becomes very slow due to the significantly larger number of channels generated by the new approach. Therefore, I added options to speed up calcChannelSurfaceOverlaps() and calcSurfaceCavitiesOverlaps() by enabling max_proc and parallel calculations.

jamesmkrieger:
sounds good

karolamik13:
@briza81 @MatthewLicht 

1. input atoms filtering from TODO list is now added

karolamik13:
I was testing diagram='weighted' in calcChannels().
I had the error1 below, but it wasn't obvious how to install vorpy. After checking, I found out that there are two versions of vorpy, but vorpy3 is needed here (that info is now added to the docs). After installation, the calculations went fine, but then I had a problem with getVmdModel() (see Error2), which I fixed by installing: python -m pip install -U dash (this is also added to the docs just in case someone would have the same problem).


Error1:
    raise ImportError('diagram="weighted" requires the vorpy package for the '
ImportError: diagram="weighted" requires the vorpy package for the additively-weighted (Apollonius) Voronoi diagram. Install vorpy, or use diagram="homogenized"/"simple".

Error2:
DeprecationWarning: The `ipykernel.comm.Comm` class has been deprecated. Please use the `comm` module instead.For creating comms, use the function `from comm import create_comm`.

karolamik13:
I tested calcSurfaceCavities()/calcSurfaceCavitiesMultipleFrames() and calcSurfaceCavityOverlaps(). With the current setups (changed in this pull request), the number of surface cavities is much larger compared to the previous setups and might be overwhelming when visualized in a graphical visualization program. I am a bit less happy with the outcome compared to the old code, at least when checking on a few examples. This part should be checked on known examples with well-defined druggable surface cavities. I was using 2KID and a few others, suggested by Issabele.

The figure below corresponds to the analysis I performed previously with the old code; see https://github.com/karolamik13/ProDy-website/blob/master/tutorials/cavitracer_tutorial/multi_modelPDB.rst (last figure). calcSurfaceCavityOverlaps() extracts only surface cavities that are present in many models, but for a single PDB model, we have quite a lot.

fig

To check other systems, feel free to use this script:
[MOD_tutorial_CAV_multi.py](https://github.com/user-attachments/files/30186451/MOD_tutorial_CAV_multi.py)

_python MOD_tutorial_CAV_multi.py '2KID' 'protein and chain A'_
for the example from the tutorial (remove _r2=1.5_ from the script to use default parameters).

karolamik13 (approved):
I tested the code and compared it with the old version as well. It looks good. 
calcSurfaceCavities()/calcSurfaceCavitiesMultipleFrames() might require additional tuning, which might be added in the next pull request after tests.

PR #2254, opened 2026-07-19
prs fixes

jamesmkrieger:
These are fixes to PRS with turbo=False independent of returning directional output


jamesmkrieger:
These fixes are going to stay with my redone directional PRS because it doesn't work without them and already has some of them

PR #2264, opened 2026-08-07
Trajectory Clustering, Statistical Analysis, and Visualization Tools 

kontheodosiadis:
The code was developed under the supervision of Prof. Karolina Mikulska-Ruminska (NCU).

## Summary 

This pull request introduces a set of tools for clustering and analyzing Molecular Dynamics trajectories. It provides functions for trajectory alignment, RMSD calculations, distance matrix generation, clustering, cluster statistical analysis, visualization, and export of representative structures and clustered trajectories. 

## Workflow
The functions follow a logical pipline: 

1. **Frame-to-Reference Analysis** 
* Align trajectory frames to a reference structure. 
* Calculate RMSD for arbitrary atom selections (e.g. ligands, proteins). 

2. **Frame-to-Frame Analysis** 
* Generate pairwise RMSD distance matrices across trajectory frames. 

3. **Clustering** 
* Apply clustering algorithms on the pairwise distance matrix. 

4. **Analysis and Visualization** 
* Calculate descriptive cluster statistics. 
* Identify representative medoid structures. 
* Generate visualizations for cluster interpretation. 

5. **Export** 
* Export representative medoid structures as `.pdb` files. 
* Export aligned cluster trajectories as `.dcd` files. 

### Clustering Algorithms 

* Bottom-up Hierarchical Clustering 
* K-Medoids Clustering 
* DBSCAN Clustering and OPTICS Reachability Plot 

Clustering algorithms support multiple backends, where needed. Optimized implementations from **scikit-learn** (or **SciPy**, where appropriate) can be used, while dependency-free implementations based on **NumPy** and the Python standard library are provided whenever practical.

### Analysis Utilities 
* Vectorized pairwise RMSD distance matrix generation  
* Medoid identification 
* Cluster summary and statistics tables 
* Cluster trajectory export (PDB/DCD) 

### Visualization 

* Hierarchical clustering dendrograms with optional cutoff visualization
* OPTICS reachability plots with optional DBSCAN $\epsilon$ threshold visualization
* RMSD histograms with optional Kernel Density Estimation (KDE) overlays 

@karolamik13  @jamesmkrieger 

[01_rmsd_frame_to_reference.ipynb](https://github.com/user-attachments/files/30826774/01_rmsd_frame_to_reference.ipynb)
[02_rmsd_frame_to_frame.ipynb](https://github.com/user-attachments/files/30826775/02_rmsd_frame_to_frame.ipynb)
[03_hierarchical_clustering.ipynb](https://github.com/user-attachments/files/30826776/03_hierarchical_clustering.ipynb)
[04_kmedoids_clustering.ipynb](https://github.com/user-attachments/files/30826777/04_kmedoids_clustering.ipynb)
[05_density_clustering.ipynb](https://github.com/user-attachments/files/30826778/05_density_clustering.ipynb)


prody/proteins/interactions.py
```diff
@@ -4237,7 +4243,1854 @@ def calcSignatureInteractions(PDB_folder, **kwargs):
...
+    >>> pdb = prody.parsePDB("structure.pdb")
+    >>> dcd = prody.Trajectory("trajectory.dcd")
+    >>> ref_coords, aligned_coords = prody.alignTrajectory(pdb, dcd, select='resname IOA')
+    >>> rmsd = prody.calcRMSDfromReference(ref_coords, aligned_coords)
+    """
+    
+    aligned_coords = np.asarray(aligned_coords)
+    reference_coords = np.asarray(reference_coords)
+    
+    if reference_coords.ndim != 2:
```
jamesmkrieger: You can support a 3-dimensional reference coords with shape (1, n_atoms, 3) and an atomic object that has a getCoords method

prody/proteins/interactions.py
```diff
@@ -4237,7 +4243,1854 @@ def calcSignatureInteractions(PDB_folder, **kwargs):
...
+        raise ValueError(f"reference_coords must be a 2D array of shape (n_atoms, 3), but got {reference_coords.shape}")
+    
+    if aligned_coords.ndim != 3:
+        raise ValueError(f"aligned_coords must have shape (n_frames, n_atoms, 3), but got {aligned_coords.shape}.")
+
+    if reference_coords.shape != aligned_coords.shape[1:]:
+        raise ValueError(f"Incompatible shapes: reference is {reference_coords.shape}, but aligned frames have {aligned_coords.shape[1:]}.")
+        
+    # Vectorized RMSD calculation
```
jamesmkrieger: prody already has rmsd calculation. Is this version quicker? If so, it should be integrated with what we already have. If not, it should be replaced here with calcRMSD or getRMSD

prody/proteins/interactions.py
```diff
@@ -4237,7 +4243,1854 @@ def calcSignatureInteractions(PDB_folder, **kwargs):
...
+    sd = (aligned_coords - reference_coords)**2
+    msd = np.mean(sd.sum(axis=2), axis=1)
+    rmsd = np.sqrt(msd)
+        
+    return rmsd
+
+
+def showRMSDfromReference(rmsd_array, title = "Frame-to-Reference RMSD",
```
jamesmkrieger: This one belongs in prody/dynamics/plotting.py and could calculate the rmsd inside like other prody show... functions. Otherwise it's not necessary to have. 

Also make sure it supports args and kwargs for other plotting options

prody/proteins/interactions.py
```diff
@@ -4237,7 +4243,1854 @@ def calcSignatureInteractions(PDB_folder, **kwargs):
...
+    # Flatten frame coordinates (n_frames, n_atoms, 3) -> (n_frames, n_atoms * 3)
+    coords_flat = aligned_coords.reshape(n_frames, 3 * n_atoms)
+    
+    euclidean_dists = cdist(coords_flat, coords_flat, metric = 'euclidean')
+    distance_matrix = euclidean_dists / np.sqrt(n_atoms)
+    
+    return distance_matrix
+
+
+def showPairwiseRMSDHeatmap(distance_matrix, title = "Pairwise RMSD Distance Matrix",
```
jamesmkrieger: again put this in prody/dynamics/plotting.py and put calculation inside. You can provide an optional argument for giving a precalculated matrix. Also, make sure you are not reinventing the wheel as we already have showMatrix

prody/proteins/interactions.py
```diff
@@ -4237,7 +4243,1854 @@ def calcSignatureInteractions(PDB_folder, **kwargs):
...
+    ax.figure.colorbar(im, ax = ax, label = label)
+
+    ax.set_xlabel("# Frame")
+    ax.set_ylabel("# Frame")
+    ax.set_title(title)
+
+    return ax
+    
+
+def showRMSDHistogram(rmsd_data, bins = 50, title = 'Distribution of RMSDs',
```
jamesmkrieger: This could be a general function to create a histogram with any data and could go in prody/utilities/catchall.py alongside showMatrix and showLines and showBars. Again make sure there are args and kwargs for greater customisation.

prody/proteins/interactions.py
```diff
@@ -4237,7 +4243,1854 @@ def calcSignatureInteractions(PDB_folder, **kwargs):
...
+    
+    if label is not None:
+        ax.legend()
+        
+    ax.grid(axis = 'y', alpha = 0.3)
+
+    return ax
+    
+
+def calcClusterPopulations(cluster_ids):
```
jamesmkrieger: This function genuinely looks novel and useful. I still need to look at this one and everything below it properly.

jamesmkrieger (changes requested):
I still need to look at the details but in general this seems like a good addition. I have two high-level comments for now.

1. move this code into different files instead of interactions.py as it is based on rmsd, not interactions. It may well get split across multiple files, but you will probably have one main one like prody/dynamics/rmsd_ clustering.py

There are currently rmsd clustering and low level plotting tools in prody/utilities/catchall.py, so that could be a possible location for some or all of it, but probably it would be better to have a dedicated file prody/dynamics/rmsd_ clustering.py that could take most of the new code. 

Plotting tools could go in either prody/dynamics/plotting.py or prody/utilities/catchall.py depending whether they are high-level (plotting particular dynamics-related features and including the calculations and plotting together) or low-level (just providing plotting utilities like histograms, which could also be quite useful).

prody/utilities/catchall.py has clusterSubfamilies and clusterMatrix that are data agnostic and take a matrix of similarities or distances/rmsds; calcRMSDclusters, which is aliased as calcGromosClusters and calcGromacsClusters, that is based on the method used on Gromacs; and calcKmedoidClusters.

2. There is some reinventing of the wheel here that needs to be consolidated with existing functions. I have commented on some examples.

prody/proteins/interactions.py
karolamik13: James, should we still avoid f-strings (.format() was usually used to support Python 2.7 users)? 
If so, we have some in the code (including the one above).

jamesmkrieger: Probably doesn't matter that much at this point as there won't really be any Python 2.7 users besides the website and we're not rebuilding it at the moment

kontheodosiadis:
I moved the files to other folders according to your suggestion. I also cleaned things up by removing unnecessary functions and utilizing existing ProDy ones where appropriate.

For now, I have kept my clustering functions separately in `rmsd_clustering.py`. While there is some overlap with existing functions, my implementation utilizes different libraries and supports multiple backends, so I think for the time-being it is safer to have them apart.

I have attached the notebooks with the corrected functions.
[01_rmsd_frame_to_reference.ipynb](https://github.com/user-attachments/files/30900503/01_rmsd_frame_to_reference.ipynb)
[02_rmsd_frame_to_frame.ipynb](https://github.com/user-attachments/files/30900504/02_rmsd_frame_to_frame.ipynb)
[03_hierarchical_clustering.ipynb](https://github.com/user-attachments/files/30900505/03_hierarchical_clustering.ipynb)
[04_kmedoids_clustering.ipynb](https://github.com/user-attachments/files/30900506/04_kmedoids_clustering.ipynb)
[05_density_clustering.ipynb](https://github.com/user-attachments/files/30900507/05_density_clustering.ipynb)

prody/dynamics/__init__.py
```diff
@@ -392,3 +392,7 @@
 from . import anmd
 from .anmd import *
 __all__.extend(anmd.__all__)
+
+from . import rmsd_clustering
+from .rmsd_clustering import *
+__all__.extend(rmsd_clustering.__all__)
```
jamesmkrieger: add back empty line

prody/dynamics/plotting.py
```diff
@@ -2703,3 +2707,304 @@ def showSelectionMatrix(matrix, atoms, selstr_x=None, selstr_y=None, **kwargs):
...
+
+        if aligned_coords.ndim != 3 or aligned_coords.shape[2] != 3:
+            raise ValueError(f"aligned_coords must have shape (n_frames, n_atoms, 3), but got {aligned_coords.shape}.")
+
+        if ref_coords.shape != aligned_coords.shape[1:]:
+            raise ValueError(f"Incompatible shapes: reference is {ref_coords.shape}, but aligned frames have {aligned_coords.shape[1:]}.")
+
+        rmsd_array = calcRMSD(ref_coords, target=aligned_coords)
+        
+    rmsd_array = np.asarray(rmsd_array)
```
jamesmkrieger: calcRMSD should already return an array from getRMSD

prody/dynamics/plotting.py
```diff
@@ -2703,3 +2707,304 @@ def showSelectionMatrix(matrix, atoms, selstr_x=None, selstr_y=None, **kwargs):
...
+    >>> plt.show()
+    """
+    
+    import matplotlib.pyplot as plt
+    try:
+        import seaborn as sns
+    except ImportError:
+        raise ImportError("The 'seaborn' package is required to display the histogram."
+                          "\nPlease install it using 'pip install seaborn'."
+                          "\nAlternatively, use standard matplotlib.pyplot.hist() for basic plots.")
```
jamesmkrieger: can the code use matplotlib hist as a fall back rather than raising an error and asking the user to plot it themselves?

prody/dynamics/plotting.py
```diff
@@ -2703,3 +2707,304 @@ def showSelectionMatrix(matrix, atoms, selstr_x=None, selstr_y=None, **kwargs):
...
+        
+    title = kwargs.pop('title', 'Intra-Cluster RMSD Distributions')
+    xlabel = kwargs.pop('xlabel', 'RMSD to Medoid [Å]')
+    ylabel = kwargs.pop('ylabel', 'Frequency')
+    grid = kwargs.pop('grid', True)
+    
+    user_label = kwargs.pop('label', None)
+
+    # Seaborn Defaults
+    kwargs.setdefault('bins', 50)
```
jamesmkrieger: can the user provide these kwargs to override these defaults e.g. if there are much fewer structures so 50 bins is too many?

prody/proteins/interactions.py
```diff
@@ -4237,7 +4237,8 @@ def calcSignatureInteractions(PDB_folder, **kwargs):
 
             # Proceed with plotting
             plot_barh(result, bond_type, n_per_plot=n_per_plot, min_height=min_height)
-
```
jamesmkrieger, line 4240: revert this minimal change so this file doesn't show up as changed

prody/utilities/catchall.py
```diff
@@ -1283,3 +1283,109 @@ def calcKmedoidClusters(coordsets, nClusters):
...
+    >>> plt.show()
+    """
+
+    import matplotlib.pyplot as plt
+    try:
+        import seaborn as sns
+    except ImportError:
+        raise ImportError("The 'seaborn' package is required to display the histogram."
+                          "\nPlease install it using 'pip install seaborn'."
+                          "\nAlternatively, use standard matplotlib.pyplot.hist() for basic plots.")
```
jamesmkrieger: again, it's better to support matplotlib as a fall back too

prody/utilities/catchall.py
```diff
@@ -1283,3 +1283,109 @@ def calcKmedoidClusters(coordsets, nClusters):
...
+    ax.set_xlabel(xlabel)
+    ax.set_ylabel(ylabel)
+    
+    if label is not None:
+        ax.legend()
+    
+    if grid:
+        ax.grid(axis='y', alpha=0.3)
+
+    return ax
```
jamesmkrieger: add back empty line

prody/dynamics/rmsd_clustering.py
```diff
@@ -0,0 +1,1491 @@
...
+            best_medoids = medoids.copy()
+
+    distances = distance_matrix[:, best_medoids]
+    cluster_ids = np.argmin(distances, axis=1) + 1 
+
+    return cluster_ids, best_medoids, best_cost
+
+
+def _clusterKMedoidsSklearn(distance_matrix, k, method_sklearn, seed, max_iter, n_init):
+    """K-Medoids clustering with sklearn_extra"""
```
jamesmkrieger, line 1129: I already have a function calcKmedoidClusters in utilities/catchall.py that uses sklearn_extra too but it's based on coords directly instead of distance matrix and doesn't have iterations. Add something int he doc string to highlight the difference and also add a docstring to mine (sorry I forgot to add one) and point to yours there too. 

I would also be good to document the parameters in the expanded docstring here and to give them default values.

prody/dynamics/rmsd_clustering.py
```diff
@@ -0,0 +1,1491 @@
...
+
+
+def clusterKMedoids(distance_matrix, k, method='alternate', method_sklearn='pam', 
+                    initial_medoids=None, seed=None, max_iter=100, n_init=10):
+    """
+    Performs K-Medoids clustering using various algorithms.
+
+    This function acts as a facade, routing the clustering task to the specified backend 
+    ('alternate', 'pam', or 'sklearn'). 'alternate' is generally faster, while 'pam' typically
+    finds solutions with lower cost. 'sklearn' needs the sklearn_extra.cluster module to be installed.
```
jamesmkrieger, line 916: I suppose it would be good to have docstrings for all the backends even though the idea isn't for users to use them directly most of the time. I made a comment on the sklearn_extra version about mentioning my related function calcKmedoidClusters in utilities/catchall.py and really that applies to this driver function more.

prody/dynamics/rmsd_clustering.py
```diff
@@ -0,0 +1,1491 @@
...
+from prody.proteins import writePDB
+from prody.trajectory import writeDCD
+from prody.measure import calcTransformation
+
+__all__ = ['alignTrajectory', 'calcPairwiseRMSD', 'calcClusterPopulations', 'getCluster', 'getClusterMedoid',
+           'calcClusterStatistics', 'calcAllClusterStatistics', 'showClusterStatisticsTable', 'writeClusters',
+           'clusterHierarchical', 'showDendrogram', 'clusterKMedoids', 'clusterDBSCAN', 'showReachabilityPlot']   
+
+
+def alignTrajectory(atoms, trajectory, align='protein and backbone', select='all'):
```
jamesmkrieger, line 24: This probably belongs in prody/trajectory instead as it appears to be specific to the Trajectory object type and not to your particular pipeline that generally takes aligned coords as an array from any type of object. Also check that the Trajectory object doesn't already have some method to do this like Ensemble has superpose. 

It would also be good to check related objects like TrajBase, TrajFile and DCDFile

prody/dynamics/rmsd_clustering.py
```diff
@@ -0,0 +1,1491 @@
...
+    extracted for every frame.
+    
+    
+    :arg atoms: reference structure used for the alignment.
+    :type atoms: :class:`prody.Atomic`
+    
+    :arg trajectory: trajectory containing the coordinate sets to align
+    :type trajectory: :class:`prody.Trajectory`
+    
+    :arg align: atom selection used to calculate the alignment transformation.
```
jamesmkrieger, line 40: include selstr or something like that in the variable name so it's immediately obvious what it is without reading the doc string

prody/dynamics/rmsd_clustering.py
jamesmkrieger: alternatively you could make this function more generic to align any type of object that resembles a trajectory and has coordsets, including ensemble and multi-model atomic objects. That could be nice to have

prody/dynamics/rmsd_clustering.py
```diff
@@ -0,0 +1,1491 @@
...
+    try:
+        for i, frame in enumerate(trajectory):
+            trans = calcTransformation(atom_align, ref_align)
+            trans.apply(atom_select)
+            aligned_coords[i] = atom_select.getCoords()
+    finally:
+        atoms.setCoords(orig_coords)
+        trajectory.reset()
+    
+    return ref_coords, aligned_coords
```
jamesmkrieger, line 96: also give an option of returning the aligned trajectory object itself

prody/dynamics/rmsd_clustering.py
```diff
@@ -0,0 +1,1491 @@
...
+
+
+def calcPairwiseRMSD(aligned_coords):
```
jamesmkrieger, line 99: It would be good to have a method like this for TrajBase like Ensemble has getRMSDs(self, pairwise=False), but I suppose that is a bigger change that we maybe don't want to make right now

prody/proteins/interactions.py
jamesmkrieger: done

jamesmkrieger:
> I moved the files to other folders according to your suggestion. I also cleaned things up by removing unnecessary functions and utilizing existing ProDy ones where appropriate.

Thanks. This looks good now.

> For now, I have kept my clustering functions separately in `rmsd_clustering.py`. While there is some overlap with existing functions, my implementation utilizes different libraries and supports multiple backends, so I think for the time-being it is safer to have them apart.

Yes, I agree. I suggest making mention of the others in the docstrings to help with users getting confused otherwise.

> I have attached the notebooks with the corrected functions

These notebooks use a filename for a file that I don't have so I won't be able to run them, but I will still glance over them seeing as the outputs are there and see if I have any more comments before finalising my review. 

Can you put a smaller version of the file in tests/datafiles (maybe with different numbers 1, 2 and 3 frames per cluster for 3 clusters) and make these into unit tests? These would also serve as illustrative examples for users of how to run the code

It would also be good to convert some of these into tutorials for the prody website at some point although I don't know whether/when we can actually update it to include them. They also serve as useful findable tutorials on the prody website repo where they also render pretty well

jamesmkrieger:
The notebooks generally look great. I have a few comments on them for improvement, but these do not generally impact the integrated code. One that does is that I see that all values in the tables in notebooks 04 and 05 including the part using showClusterStatisticsTable are represented as floats with several decimal places. I'd generally reduce actual floats to have 2 or 3 decimal places and convert those that actually aren't floats at all like Total Frames and Medoid Frame to int.

- The approach comparison results in notebook 01 with results reproducing to residuals within 10^-7 to 10^-8 or 10^-15 are a perfect basis for being able to compare against a reference array from one of these calculations with numpy allclose like you have in notebook 02, which could also be a unit test. 

- Note the rmsd 0 at frame 1003 is just because that is the start of run 3 and matches the start of run 2 and the first residuals also reflect that. 

- Based on notebooks 01 and 02, it is clear that there are two clear clusters in run 2 that look like they could well be populated in trajectory 3 too, just with much shorter residence times and more interconversion. Can you map out which frames end up in which clusters and whether it matches the expectations? 

- Does the outlier cluster of 9 frames correspond to the starts of both runs that then don't persist in the hierachical case in notebook 03 just like these frames 0,1,2,1003 show up in the DBSCAN noise frames in notebook 05? The 1D RMSD plots in notebook 01 do show minimum RMSDs from the closer cluster in the range of 3.8 so they could well do. The cutoff is a bit arbitrary really but I'd say 3.8 removing these outlier frames is good.

- It would be good to look at what's different about frames 605, 854 and 855 that make them outliers too. 605 looks like it's a brief deviation from the cluster in that time period. Perhaps 854 and 855 are at the transition between that cluster and going back into the other cluster at the end of run 2. 

- There's also a typo in the markdown cell underneath code cell 8 that prints the frames detected as noise and in a few other places. You should spell separated with an a after the p. There are some other typos around too

- The number of clusters colour bar doesn't make sense. There should be fewer colours to match the matrix. You can truncate the tab10 cmap to just cover a smaller range of values

- I think you worded it a bit negatively that DBSCAN tended to prefer one cluster but you justified it quite well. It's a bit surprising that your final result gives a much smaller cluster 2 than the other methods, so probably you want to comment on that too. I'd maybe pick epsilon 1.4 or 1.5 as that should split the two main clusters instead of just splitting off a minor one. You could take one of the cases with 3 clusters so the minor one is also captured and I'd say the high noise is a real biological noise that you don't want to optimise away.

prody/dynamics/rmsd_clustering.py
```diff
@@ -0,0 +1,1491 @@
...
+    for key, label in metrics:
+        row_labels.append(label)
+        table.append([stat[key] for stat in all_stats])
+        
+    if not isinstance(show, bool):
+        raise TypeError("show must be a bool.")
+    
+    if show:
+        kwargs.setdefault('tablefmt', 'fancy_grid')
+        kwargs.setdefault('floatfmt', '.4f')
```
jamesmkrieger: This default isn't right for all rows. Some of them are truly ints and most of the others don't deserve so much precision either.

jamesmkrieger (changes requested):
Here are a fair few changes, but it's looking really good now.

prody/dynamics/plotting.py
kontheodosiadis: I have `np.asarray` there to sanitize direct `rmsd_array` input from the user. And since it was already there, I thought it wouldn't hurt to do it regardless of the input. However, to make its purpose clearer I have now put it under `else`, so now it doesn't run redundantly after `calcRMSD`.

prody/dynamics/plotting.py
kontheodosiadis: Yes, they absolutely can! `kwargs.setdefault()` sets 50 as the default, but if the user passes `bins=10` it will simply override it. The same goes for any other parameter from `seaborn.histplot`.

prody/dynamics/rmsd_clustering.py
kontheodosiadis: I’ve updated the public `.. seealso::` sections to cross-reference `clusterMatrix` with `clusterHierarchical`, and `calcKmedoidClusters` with `clusterKMedoids`. For the private helper functions, I left out parameter docstrings and default argument values, since all parameter definitions and default resolution are handled at the public facade level, so duplicating them internally would be redundant.

prody/dynamics/rmsd_clustering.py
kontheodosiadis: For the private helper functions, I left out parameter docstrings and default argument values, since all parameter definitions and default resolution are handled at the public facade level, so duplicating them internally would be redundant. I have updated the public `:seealso::` section to cross-reference them.

prody/dynamics/rmsd_clustering.py
```diff
@@ -367,22 +367,25 @@ def calcAllClusterStatistics(distance_matrix, cluster_ids):
     Returns a list of dictionaries, each corresponding to a cluster with the statistic quantities
     as keys.
     
-    
     :arg distance_matrix: a two-dimensional square, symmetric pairwise distance matrix.
     :type distance_matrix: :class:`numpy.ndarray`
     
-    :arg cluster_ids: a one-dimensional array of the cluster IDs per element
+    :arg cluster_ids: a one-dimensional array of 1-indexed cluster IDs per element.
```
jamesmkrieger, line 373: you could increment all values to start at 1 and warn rather than skipping them

jamesmkrieger (approved):
I think we're more or less in a good state now

karolamik13 (approved):
I tested all the functions. Working great.

PR #2258, opened 2026-07-21
ClustENM: multi-start (setAtoms accepts a list of AtomGroups) + optional fixer

jamesmkrieger:
setAtoms now accepts a list/tuple of AtomGroups to seed a MULTI-START run: the topology is built from the first structure and a matching coordinate set is collected from each (via _fix_multi, same PDBFixer processing so atom order matches). They become the generation-0 initial population -- each minimised -- so a run can start from a SET of structures rather than one; equivalent to skipping the single-conformer gen-0. The generation loop is unchanged and then samples normal modes from all of them. At n_gens=0 this is a batched relax that keeps the originals (no clustering).

setAtoms(..., fix=True) gains fix=False: skip PDBFixer and build the OpenMM topology directly from atoms as-is (only valid when already complete + protonated; forcefield build fails otherwise). Added _nofix for that path. Closes #2089 

Validated with distorted structures from boltz: setAtoms([s0,s1]) + run(n_gens=0) minimises both (clash 318/340 -> 0, domain distance and %SS preserved). Single-structure path byte-unchanged.

PR #2269, opened 2026-08-12
ANMD and ClustENM fixes and test fixes

jamesmkrieger:
Also includes improved checker for RTB blocks

This should fix the ANMD test problems mentioned in #2264 - @kontheodosiadis please confirm

This is also a prerequisite that more clustenm tests for #2258 will be built upon


jamesmkrieger:
@karolamik13 all tests pass including ones for clustenm multi-start

PR #2274, opened 2026-09-04
CaviTracer more robust automatization and removal of similar channel alternatives

briza81:
Making the calculations and its default robust wrt to different structures, limiting the need for specific parameter adjustments. Also, similar tunnel variants are better pruned.

Measured over the 21-structure testing corpus (MOLE2 + CAVER reference sets, a porin and two protonated dehalogenases), each version with its own defaults:

| | `main` (32bd69e7) | this branch |
|---|---|---|
| channels reported, whole corpus | 278 | 147 |
| widest bottleneck | - | equal or wider in 19 of 21 cases (2fgq 0.83 -> 2.20 Å, 2bg9 1.98 -> 7.64 Å); 1grm 1.43 -> 1.34 and 2ace 2.00 -> 1.99 are the exceptions |
| corpus wall time | 206.5 s | 148.5 s (1.39x; 1m56 alone 62.4 -> 16.9 s) |

Eighteen of the 21 cases report fewer channels while the widest bottleneck holds or improves. **Two separate things cause that**, and the companion document separates them by running `main`'s own algorithm a third time at the branch's `inner_radius` and `surf_radius`: 278 -> 212 -> 147, so the recalibrated defaults account for -66 and the deduplication for -65.  The two cases that report *more* are the wide-pore ones, from `surf_radius`.

## Major changes

- **One channel per way out.** A surface opening is now a mouth's inscribed ball instead of the tetrahedron at its centre - mouths are one tetrahedron thick,  so a route could walk around one and surface again a few Angstrom away, and
  the two were then compared as distinct exits. A search stops at the first opening it enters, the two mouths of a pair are compared symmetrically, and duplicates are merged on how far they *diverge* in Angstrom (scaled by the local clearance = surface distance) . When two routes merge, the better ending is reported - a wider mouth or a wider bottleneck, with neither bought by narrowing the other.
- **Seeding from a user start point.** The seed's depth floor was the anchor's own depth, but a `start_point` is usually placed on a ligand or a catalytic residue, deeper than the widest part of its pocket - so nothing was eligible,  the seed stayed on the narrow cell nearest the point, and every channel of that site was silently capped at its radius (3eyx reported 1.30 Å twice; the routes are 1.78 and 1.65 Å). Every automatic seed is now printed as a pasteable `start_point`.
- **Centerline in Angstrom.** The spline was parameterized by tetrahedron index, and tetrahedra are not evenly spaced, so it spent equal parameter on unequal distances. The two curves never part by more than 0.36 Å anywhere in the corpus - the route is the same - but the old one reverses direction between consecutive samples, up to 178°, and that zig-zag inflated every reported length by 1.8-9.9 % per case (17.8 % on one channel). It is now centripetal in distance, with a PCHIP radius profile.
- **Protonation-aware input check.** The old test asked only whether any hydrogen was present, so a structure with polar hydrogens only passed as protonated and could be probed below water size. Hydrogens per heavy atom are now scored per chain type (protein ~1.0, nucleic acid ~0.55); a mixed structure counts as protonated only when every chain type in it is, and the message names whichever is short. Nucleic acid is named in the composition report.

## Altered defaults

| | `main` | this branch |
|---|---|---|
| `surf_radius` | 3 | **15** |
| `inner_radius` | 0.9 | **1.2** - the smallest probe giving similar channels with and without explicit hydrogens |
| `sparsity` | 1 | **6**, now in Angstrom: the smallest centre-to-centre distance at which two openings count as separate |
| `similarity`, `route_tolerance` | 0.8, 1.0 | **replaced by** `route_divergence=0.2` |

### DIFF
--- a/prody/proteins/channels.py
+++ b/prody/proteins/channels.py
@@ -608,12 +608,13 @@ def showChannels(channels, model=None, surface=None):
     conda install open3d (for Anaconda users; version open3d-0.19.0 was used 
     during the development) or pip install open3d
     
-    :arg channels: A list of channel objects or a single channel object. Each 
-        channel should have a `getSplines()` method that returns two 
-        CubicSpline objects: one for the centerline and one for the radii.
+    :arg channels: A list of channel objects or a single channel object. Each
+        channel should have a `getSplines()` method that returns two
+        interpolators over one parameter domain: one for the centerline and one
+        for the radii.
     :type channels: list or single channel object
-    
-    :arg model: An optional Open3D TriangleMesh object representing the 
+
+    :arg model: An optional Open3D TriangleMesh object representing the
         molecular model, such as a protein. If provided, this model will be 
         rendered in the visualization.
         Model can be generated using getVmdModel() function.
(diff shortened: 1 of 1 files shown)


briza81:
@karolamik13 pls heave a look, attaching also a bit more detailed descriptions
[details.pdf](https://github.com/user-attachments/files/31841876/details.pdf)

karolamik13 (approved):
Everything looks good.

I tested systems that are described in the ProDy-website tutorial for CaviTracer and some additional systems, which will be used as a case study.

I have to say that the overlapping surface calculations (for multiple conformations) are much better, the occupancy values are much higher, and this is because channels that had similar exits before were counted a few times.

PR #2279, opened 2026-09-17
VMD visualization of CaviTracer

karolamik13:
I created the writeVmdCaviTracerScript() function, which creates a VMD state for the results obtained by CaviTracer.

How to use it:
from prody import *
p = parsePDB('1tqn')
atoms = p.select("protein")

# For channels:
channels, surface = calcChannels(atoms)
writeVmdCaviTracerScript(channels, atoms)

# Surface cavities:
cavities, surface = calcSurfaceCavities(atoms)
writeVmdCaviTracerScript(cavities, atoms, object_type='surface_cavities', surface=surface)

It will also work for pores and reconstruction of channels+surface cavities.

The user is obtaining a TCL script (vis_channels.tcl for channels, vis_pores.tcl for pores, etc.) that contains a visualization-ready state of the analysis.

To run the script:
vmd -e vis_channels.tcl

Visualization:
cavitracer_figure41


prody/proteins/channels.py
```diff
@@ -10905,3 +10905,425 @@ def _writeVisScript(directory, pattern='chl*.pqr'):
...
+                "surface returned by calcSurfaceCavities() must be provided "
+                "for object_type='{0}'.".format(object_type))
+
+    if objects is None:
+        raise ValueError("objects cannot be None")
+
+    if object_type == 'connected_cavities_channels':
+        if isinstance(objects, dict):
+            objects = [objects]
+    elif not isinstance(objects, list):
```
jamesmkrieger: This is redundant. The same happens in both if branches

prody/proteins/channels.py
```diff
@@ -10905,3 +10905,425 @@ def _writeVisScript(directory, pattern='chl*.pqr'):
...
+        objects = [objects]
+
+    if len(objects) == 0:
```
jamesmkrieger: This will never happen unless an empty list is explicitly provided because of wrapping into a list giving length 1

prody/proteins/channels.py
```diff
@@ -10905,3 +10905,425 @@ def _writeVisScript(directory, pattern='chl*.pqr'):
...
+        if not (result_file.exists() or
+                protein_file.exists() or
+                script_file.exists()):
+            break
+
+        number += 1
+
+    # ------------------------------------------------------------------
+    # Save protein
+    # ------------------------------------------------------------------
```
jamesmkrieger: Remove these as there are no long blocks that need separating

jamesmkrieger: For the other ones they are fine but could be shorter

prody/proteins/channels.py
```diff
@@ -10905,3 +10905,425 @@ def _writeVisScript(directory, pattern='chl*.pqr'):
...
+
+set protein_file [file join $script_dir {__PROTEIN_FILE__}]
+set result_file  [file join $script_dir {__RESULT_FILE__}]
+set object_type "__OBJECT_TYPE__"
+
+# ----------------------------------------------------------------------
+# Display
+# ----------------------------------------------------------------------
+
+color Display Background white
```
jamesmkrieger, line 11127: Maybe make these arguments

jamesmkrieger (approved):
A few minor things that I don’t need to check on. Otherwise everything looks good

